PR #2, opened 2015-06-30
Fixed lowercase b of LoopBack in the example

MaxBosse:
Using consumer.Loopback will cause in a type-error when starting gollum. Actual name of the consumer is consumer.LoopBack


andygrunwald:
Thx

PR #11, opened 2015-12-29
Add RPM spec file for building on RHEL7

gmarkey:
Tested on CentOS 7.1


andygrunwald:
@gmarkey Thank you. We will have a detailed look into this starting next year, ok? (because in germany we got public holiday).

@arnecls Can you take over?

arnecls:
Sure. One thing I have to note is that /Contrib was meant to be used for other things. Maybe we should create a new folder for these kind of things.
I will wrap my head around this in the next days.

arnecls:
The files can now be found in /dist

PR #19, opened 2016-05-16
fix kinesis consumer to respect DefaultOffset

relud:
Currently if `DefaultOffest` configuration is set to `newest` or `oldest`, and `OffsetFile` points to a file that does not contain all of the current kinesis shards then:
<pre>
iterator, err := cons.client.GetShardIterator(&iteratorConfig)
</pre>
will fail, with the following error:
<pre>
InvalidArgumentException: Must either specify (1) AT_SEQUENCE_NUMBER or AFTER_SEQUENCE_NUMBER and StartingSequenceNumber or (2) TRIM_HORIZON or LATEST and no StartingSequenceNumber. Request specified AFTER_SEQUENCE_NUMBER and no StartingSequenceNumber.
</pre>

With this patch the `DefaultOffset` will be respected on new shards, while shards found in the `OffsetFile` will use `ShardIteratorTypeAfterSequenceNumber`.

### DIFF
--- a/consumer/kinesis.go
+++ b/consumer/kinesis.go
@@ -192,7 +192,6 @@ func (cons *Kinesis) Configure(conf core.PluginConfig) error {
 		if err != nil {
 			return err
 		}
-		cons.offsetType = kinesis.ShardIteratorTypeAfterSequenceNumber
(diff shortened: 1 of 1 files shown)


arnecls:
Good find! 👍

PR #20, opened 2016-05-17
write OffsetFile in kinesis consumer

relud:
previously the kinesis consumer never actually wrote to `OffsetFile`, it would only read from it. with this patch, the kinesis consumer will write to `OffsetFile` every time it updates offsets.


arnecls:
Another good find :+1:

PR #21, opened 2016-05-17
fix kinesis producer to enforce BatchMaxRequests

relud:
currently the `transformMessages` function assumes that `len(messages)` is less than `BatchMaxMessages`. If that isn't the case, then it may send more than `BatchMaxMessages` to `kinesis.PutRecords`, which will reject calls with more than 500 records.

with this patch the `transformMessages` function will start a new batch request after reaching `BatchMaxMessages`.

I renamed `records` to `buffer` in order to keep naming consistent with comments, and to make a sensible name for the list of records.

### DIFF
--- a/producer/kinesis.go
+++ b/producer/kinesis.go
@@ -204,16 +204,16 @@ func (prod *Kinesis) dropMessages(messages []core.Message) {
 }
 
 func (prod *Kinesis) transformMessages(messages []core.Message) {
-	streamRecords := make(map[core.MessageStreamID]*streamData)
+	streamBuffers := make(map[core.MessageStreamID][]*streamData)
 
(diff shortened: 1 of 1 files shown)


arnecls:
As transformMessages is backed by a core.MessageBatch and the size of that batch is bound to BatchMaxMessages len(messages) will always be <= BatchMaxMessages.
See https://github.com/trivago/gollum/blob/master/core/messagebatch.go#L98

That's why there is no check in transformMessages.

arnecls:
Regarding your changes - it would probably be a better idea to simply send the batch right after the limit is detected, reset the buffer and continue. This would be more memory friendly and require less changes.

PR #22, opened 2016-05-18
allow kinesis records to contain multiple messages

relud:
currently the kinesis producer creates a kinesis record for each gollum message, and the consumer creates a gollum message for each kinesis record.

With this patch, the producer can optionally join multiple gollum messages into a single kinesis record, and the consumer can split kinesis records into multiple gollum messages.

Aggregating messages like this is desirable because records are rounded up to the next 25KB when determining cost and each shard in a stream is limited to 1K record/s or 1MB/s.

### DIFF
--- a/producer/kinesis.go
+++ b/producer/kinesis.go
@@ -50,6 +50,8 @@ const (
 //    CredentialFile: ""
 //    CredentialProfile: ""
 //    BatchMaxMessages: 500
+//    RecordMaxMessages: 1
+//    RecordMessageDelimiter: "\n"
 //    SendTimeframeSec: 1
 //    BatchTimeoutSec: 3
 //    StreamMapping:
@@ -76,6 +78,12 @@ const (
 // BatchMaxMessages defines the maximum number of messages to send per
 // batch. By default this is set to 500.
 //
+// RecordMaxMessages defines the maximum number of messages to join into
+// a kinesis record. By default this is set to 500.
+//
+// RecordMessageDelimiter defines the string to delimit messages within
+// a kinesis record. By default this is set to "\n".
+//
 // SendTimeframeMs defines the timeframe in milliseconds in which a second
 // batch send can be triggered. By default this is set to 1000, i.e. one
 // send operation per second.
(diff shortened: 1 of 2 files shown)


producer/kinesis.go
```diff
@@ -123,6 +134,8 @@ func (prod *Kinesis) Configure(conf core.PluginConfig) error {
 
 	prod.streamMap = conf.GetStreamMap("StreamMapping", "default")
 	prod.batch = core.NewMessageBatch(conf.GetInt("BatchMaxMessages", 500))
+	prod.recordMaxMessages = conf.GetInt("RecordMaxMessages", 1)
+	prod.delimiter = []byte(conf.GetString("RecordMessageDelimiter", "\n"))
```
arnecls: default value should be "" to match consumer and make old behavior default

producer/kinesis.go
```diff
@@ -123,6 +134,8 @@ func (prod *Kinesis) Configure(conf core.PluginConfig) error {
 
 	prod.streamMap = conf.GetStreamMap("StreamMapping", "default")
 	prod.batch = core.NewMessageBatch(conf.GetInt("BatchMaxMessages", 500))
+	prod.recordMaxMessages = conf.GetInt("RecordMaxMessages", 1)
```
arnecls: Give an error if RecordMaxMessages is == 0 or force set it to shared.MaxI(1,conf.GetInt("RecordMaxMessages", 1)).
Otherwise there could be a divide by 0 later.

producer/kinesis.go
relud: The old behavior is default due to RecordMaxMessages defaulting to 1. Will
change this to "" to match consumer. Perhaps I should log a warning if
RecordMaxMessages > 1 and this is still ""? because the consumer can't
split on "".

producer/kinesis.go
relud: I'll go with error.

producer/kinesis.go
arnecls: I'm currently checking the rest of your code if MaxMessages is switching between old and new, a default value is probably better here (so keep "\n").
As an Idea: I normally try to fix as many errors as possible without the need for a hard error, so maybe it should go like this: 
- if max messages is < 1 set it to 1 and give a warning
- if max message is > 1 and delimiter is "" set it to "\n" and give a warning

relud: Doing this:

* if max messages is < 1 set it to 1 and give a warning
* if max message is > 1 and delimiter is "" set it to "\n" and give a warning

producer/kinesis.go
relud: went with setting to 1 and throwing warning, as per comments on delimiter

arnecls:
Rest looks ok for me.

PR #25, opened 2016-05-26
kinesis producer allow * in StreamMapping

relud:
currently `StreamMapping` requires that every stream be listed explicitly, or it will default to `streamName`

with this patch, `StreamMapping` will allow for a default stream other than `streamName` by specifying something like `StreamMapping: {"*": "default"}`

this matches the documentation at the bottom of: http://gollum.readthedocs.io/en/latest/producers/kinesis.html
the default config value was changed to ensure previous default bahavior.

### DIFF
--- a/producer/kinesis.go
+++ b/producer/kinesis.go
@@ -132,7 +132,7 @@ func (prod *Kinesis) Configure(conf core.PluginConfig) error {
 	}
 	prod.SetStopCallback(prod.close)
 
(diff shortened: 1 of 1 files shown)


arnecls:
Sorry this took so long.
I will probably extend this a little more as we added this functionality to kafka, too.

relud:
cool! :)

PR #26, opened 2016-06-24
support reading unix timestamp dates in the JSON formatter

relud:


arnecls:
Looks fine to me, I would like to discuss two things though:

1. I'm not very happy with two different kind of formats in the same field without any indicator which one to use. If you want to add a new one you have to change 2 places in code and there might be collisions.
Maybe you have an idea / preference here, I can only think of prefixing or an additional field - both have their downsides.

2. Why use two notations for each format? I guess the short form is totally sufficient here.

relud:
2. removed long form

relud:
1. I'll make it a new field

relud:
Added a new field named `JSONUnixTimestampRead` for reading unix timestamps

arnecls:
:+1:

PR #27, opened 2016-06-27
Add option to check for new shards in Kinesis consumer

relud:


consumer/kinesis.go
```diff
@@ -320,6 +323,38 @@ func (cons *Kinesis) connect() error {
 		Log.Debug.Printf("Starting kinesis consumer for %s:%s", cons.stream, shardID)
 		go cons.processShard(shardID)
 	}
+
+	if cons.shardTime {
+		return nil
+	}
+
+	for cons.running {
```
arnecls: Bad idea, as line 374 will never be called.
Suggestion: Use time.AfterFunc for this

relud: doh >_<
fixed.

arnecls:
:+1: 
One last critique - the new go routine should be added to workers as we need to wait for it when shutting down. This is so minor I will add this after merge.

PR #28, opened 2016-06-27
add ProcessJSON directive to omit fields

relud:


format/processjson.go
```diff
@@ -158,6 +158,9 @@ func (values *valueMap) processDirective(directive transformDirective) {
 			case numParameters == 1:
 				(*values)[directive.key] = strings.Trim(value, shared.Unescape(directive.parameters[0]))
 			}
+
+		case "omit":
```
arnecls: I guess "remove" is a bit clearer for non-native speakers (I have colleagues here which don't even now this word ^_^)

relud: sure thing, fixed

arnecls:
:+1:

PR #31, opened 2016-06-28
Modify ProcessJSON to remove restriction to type map[string]string

relud:


arnecls:
Could you change this to use shared.MarshalMap?
This type already takes care of type based constraints like this, so you can e.g. do valueMap.GetString().
Doing so would add this to our "maintenance chain" for type based marshaling.

relud:
nevermind, I figured out a better path

relud:
updated to use shared.MarshalMap instead of valueMap

arnecls:
:+1:

PR #32, opened 2016-07-29
add aws s3 producer

relud:
copied the kinesis producer, and modified it to produce s3 objects. vendored aws-sdk-go/services/s3


relud:
wait, my testing was incomplete, this doesn't work yet

PR #38, opened 2016-08-19
add aws s3 producer

relud:


producer/s3.go
```diff
@@ -0,0 +1,349 @@
...
+)
+
+// S3 producer plugin
+// This producer sends data to an AWS S3 Bucket.
+// Configuration example
+//
+//  - "producer.S3":
+//    Region: "eu-west-1"
+//    Endpoint: "s3-eu-west-1.amazonaws.com"
+//    StorageClass: "STANDARD"
```
arnecls: Please add a more detailed documentation for StorageClass below

producer/s3.go
```diff
@@ -0,0 +1,349 @@
...
+	if sleepDuration > 0 {
+		time.Sleep(sleepDuration)
+	}
+
+
+	// Send to S3
+	for _, objects := range streamObjects {
+		for idx, object := range objects.objects {
+			hash := sha1.Sum(object)
+			key := objects.s3Prefix + time.Now().Format(prod.timeWrite) + hex.EncodeToString(hash[:])
```
arnecls: The key should be derivable from the message, e.g. by having a special formatter.
Redis, kafka (and I think kinesis, too) already do this to allow additional flexibility here.

producer/s3.go
```diff
@@ -0,0 +1,349 @@
...
+//
+// BatchTimeoutSec defines the number of seconds after which a batch is
+// flushed automatically. By default this is set to 30.
+//
+// TimestampWrite defines the go timestamp format that will be used in naming
+// objects. Objects are named <s3_path><timestamp><sha1>. By default timestamp
+// is set to "2006-01-02T15:04:05".
+//
+// StreamMapping defines a translation from gollum stream to s3 path. If
+// no mapping is given the gollum stream name is used as s3 path.
```
arnecls: You should note that the path is not "s3://bucket/path" but only "bucket/path" or "bucket".
It does make sense to change the example above to "*" : "bucket/path", too.

producer/s3.go
```diff
@@ -0,0 +1,349 @@
...
+	// Send to S3
+	for _, objects := range streamObjects {
+		for idx, object := range objects.objects {
+			hash := sha1.Sum(object)
+			key := objects.s3Prefix + time.Now().Format(prod.timeWrite) + hex.EncodeToString(hash[:])
+			params := &s3.PutObjectInput{
+				Bucket:       aws.String(objects.s3Bucket),
+				Key:          aws.String(key),
+				Body:         bytes.NewReader(object),
+				StorageClass: aws.String(prod.storageClass),
```
arnecls: ContentType, ContentEncoding and Expires might also be useful fields here.
(But we could also add them later)

relud: I would prefer to add those later

relud:
This will need follow PRs from me to add:

- file buffer, for shipping files in excess of 1GB
- compression

arnecls:
I'll play around with this a little. For a first shot this is definitely good, but my gut-feeling tells me that this needs to be a bit more flexible (as stated above).
I'm perfectly fine with doing an iterative approach here.

andygrunwald:
Should we mention the S3 producer in the README as well?

arnecls:
done

PR #39, opened 2016-08-19
add systemd consumer

relud:
Only builds with `-tags systemd`, and requires `cgo` and `systemd/sd-journal.h` to build successfully. Would've picked a pure-go reader for the systemd journal, but one doesn't seem to exist.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ before_install:
 - wget http://launchpadlibrarian.net/234454185/librdkafka-dev_0.8.6-1.1_amd64.deb
 - sudo dpkg -i librdkafka1_0.8.6-1.1_amd64.deb
 - sudo dpkg -i librdkafka-dev_0.8.6-1.1_amd64.deb
+- sudo apt-get install -y libsystemd-journal-dev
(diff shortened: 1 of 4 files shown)


makefile
```diff
@@ -35,6 +35,10 @@ aws:
 	@GOOS=linux GOARCH=amd64 $(BUILD_FLAGS) go build -o gollum
 	@zip -j dist/gollum_aws.zip gollum config/kinesis.conf dist/Procfile
 
+systemd:
```
arnecls: Please move linker / build options into the consumer file.
See https://github.com/trivago/gollum/blob/master/contrib/native/librdkafka/client.go#L17

arnecls:
Plugins using native bindings should go to contrib/native.
Please move the consumer there.

relud:
Will do

relud:
moved to contrib/native

arnecls:
Thanks a lot (as usual ^^).
:+1:

andygrunwald:
We should mention the systemd consumer as well in the README

arnecls:
done

Issue #41, opened 2016-09-08
ENTRYPOINT in Dockerfile is wrong.


breadmj:
running `docker run -it --rm trivago/gollum`
results in `docker: Error response from daemon: oci runtime error: exec: "/go/src/gollum/gollum": stat /go/src/gollum/gollum: no such file or directory.`

**current**
ENTRYPOINT ["/go/src/gollum/gollum"]

**should be**
ENTRYPOINT ["/go/src/github.com/trivago/gollum/gollum"]

or how about '*go install*' after '*go build*'?

arnecls:
The dockerfile is quite old, looks like it needs an update.
Thanks for pointing this out.

Exeu:
here is the fix #42

arnecls:
Thanks a lot!

PR #42, opened 2016-09-12
Fix Dockerfile to be able to build gollum image again.

Exeu:
The Dockerfile is quite old. So it needs to get updated a bit.
This change includes the correct gollum path and introduced the filepermission change to be able to execute the gollum binary.


arnecls:
This fixes issue #41 .
Thanks a lot!

PR #43, opened 2016-09-16
add support for ssl and sasl to native kafka producer

relud:


arnecls:
Looks fine to me :+1:

Issue #17, opened 2016-05-03
kafka consumer should support consumer groups


predmond:
Sarama cluster has support for consumer groups: https://github.com/bsm/sarama-cluster

Note sarama-cluster requires kafka >= 0.9.

arnecls:
As of 0.4.4 we included Sarama 1.10 so we finally have support for Kafka 0.10.
I think this will be added once we switch our cluster to 0.10, but if anyone wants to add this to the 0.4.5 branch I would be very grateful, too, as I am currently working on finishing the 0.5.0 changes.

arnecls:
@predmond Is this ok for you?

arnecls:
And a big thanks to @relud by the way.

PR #44, opened 2016-09-17
Kafka sarama ssl

relud:

PR #45, opened 2016-09-19
add KafkaGroup consumer

relud:
for #17

I need to test this more tomorrow.

### DIFF
--- a/consumer/kafka.go
+++ b/consumer/kafka.go
@@ -15,8 +15,12 @@
 package consumer
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
+	"fmt"
 	kafka "github.com/Shopify/sarama"
+	cluster "github.com/bsm/sarama-cluster"
 	"github.com/trivago/gollum/core"
 	"github.com/trivago/gollum/core/log"
 	"github.com/trivago/gollum/shared"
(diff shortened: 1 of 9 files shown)


arnecls:
I'm currently scratching my head on how to improve that this is almost the same as the "normal" sarama producer. At the end this stuff should be possible with the standard sarama consumer, too ... at least from the concept behind it.
I'm a bit irritated that Shopify made this a separate package.

relud:
after doing, I think I can see how to integrate this with the regular kafka consumer, as an option.

arnecls:
I guess that would be better. This looks like we end up maintaining both code bases and based on my previous experience with sarama this is something that we definitely don't want to do ;)

relud:
This is building properly, but it does not work at the moment.

relud:
This is working for me when I run it with and without a consumer group set

andygrunwald:
Awesome work @relud!

arnecls:
Am I correct that this includes PR #44 ?
I will run this through our integration test suite tomorrow.

arnecls:
integration tests 🆗

relud:
no, pr #44 is for the kafka producer.

arnecls:
stupid me.
So this will add groups AND authentication to the consumer.

relud:
yes

PR #46, opened 2016-09-20
move DayBasedIndex handling to after counter

relud:
prevents panic in elasticsearch producer.

`atomic.AddInt64(prod.counters[index], 1)` only works if index hasn't had the date appended yet.


arnecls:
It will however lead to indexes not being separated by date ... which is probably a good idea anyway. Good catch!

PR #47, opened 2016-09-27
add sample filter

relud:
This filter will accept `SampleRatePerGroup` out of every `SampleGroupSize` messages.

### DIFF
--- a/filter/sample.go
+++ b/filter/sample.go
@@ -1,0 +1,108 @@
+// Copyright 2015-2016 trivago GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


arnecls:
So this is basically filter.Rate but binds all streams going through the filter to the same counter, correct? So when checking streams A and B filter.Rate will block only A if A goes over the limit. Your filter will block A and B if either A or B go over the limit, correct?

relud:
while it is true that this doesn't filter per-stream, that's a side-effect of implementation. The desired effect is to rate limit as 2 of every 5 messages accept, instead of 2 messages per second accept.

arnecls:
Ah, I see. Maybe this can be added to filter.Rate instead?
In that case you would get the per-stream rate automatically.

arnecls:
On second thought ... nope - different usecase.

arnecls:
I would like to the the "per stream count" extension in the future, otherwise fine.

PR #48, opened 2016-09-27
add filter.Any

relud:
This filter allows multiple filters to be specified and accept if only one filter accepts. This is to allow behavior like rate limiting messages except if they start with `ERROR`, without rerouting messages that start with `ERROR`.

### DIFF
--- a/filter/any.go
+++ b/filter/any.go
@@ -1,0 +1,70 @@
+// Copyright 2015-2016 trivago GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	 http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package filter
+
+import (
+    "fmt"
+	"github.com/trivago/gollum/core"
+	"github.com/trivago/gollum/shared"
+)
+
+// Any filter plugin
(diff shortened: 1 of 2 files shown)


arnecls:
This is a good workaround for the deficits of version 0.4.x.
In 0.5.x however this will be obsolete as filters and formatters are getting merged and you can apply them in any order and as often as you want to. As of this I have to discuss this change a bit because adding an already deprecated feature feels a bit weird.

relud:
To clarify, in 0.5.x, you can `OR` modulators together?

relud:
also, if this is being fixed in 0.5.0, i'm okay with not merging it.

arnecls:
I would leave this open for those who need this in 0.4.x right now - nothing wrong with the change IMO.

In 0.5.x something like this
```
- Formatter: format.Envelope
- EnvelopeFormatter: format.Base64Decode
- Filter: filter.RegExp
- FilterAfterFormat: filter.Rate
```
Will be replaced by
```
- Modulators:
   - filter.RegExp
   - format.Base64Decode
   - format.Envelope
   - filter.Rate
```
Which is much more convenient and kills of a lot of workarounds.
It is already half way implemented in the 0.5.x branch but there is still a lot of work to do.
I have not added this to https://github.com/trivago/gollum/wiki/Breaking050, yet, because it is not 100% done.

relud:
okay. I still don't quite understand how in 0.5.x to get a message to accept if, for instance, it's is either valid JSON or matches regexp '^ERROR'.

arnecls:
Ah, now I see. 0.5.x uses AND to combine filters as modulators (filters/formatters) are called in array order with early out. A nested filter like yours would solve the issue but it still feels like this should be added to the core. On the other hand I don't know of a good way to included that, yet.
```
- Modulators:
  - filter.Any
    - filter.RegExp
    - filter.Rate
  - format.Timestamp
```
looks okay-ish to me ... could lead to duplicated code though ... maybe.
I will have a look at that.

arnecls:
Needs modification for 0.5.0

PR #49, opened 2016-09-27
add processjson directive to flatten values into new fields

relud:
so that `'Fields:flatten'` converts `{"Fields":{"message":"error"}}` to `{"Fields.message":"error"}`

and

so that `'messages:flatten:_'` converts `{"messages":["error1", "error2"]}` to `{"messages_0":"error1","messages_1:"error2"}`


arnecls:
Looks good to me

PR #53, opened 2016-11-04
add file buffer and gzip to producer.S3

relud:
This ended up pretty much being a re-write of producer.S3, but I made sure it's backwards compatible and doesn't change output behavior unless new configuration options are used.


arnecls:
Finally found some time to review this thing :+1:

PR #58, opened 2016-12-02
use journal.Wait() to notice journal rotations

### LINKED ISSUE #57: native.Systemdconsumer stops on system log rotation
the implementation of native.Systemdconsumer only calls [sd_journal_open](https://www.freedesktop.org/software/systemd/man/sd_journal_open.html) once, and the resulting journal object does not pick up new journal files after rotation, so after the systemd log rotates the consumer halts without any errors.

relud:
Fixes #57


arnecls:
Thanks again.
Sorry for letting you wait this long.
Your other merge requests will be reviewed in the following days.

PR #63, opened 2017-01-17
http basic auth

glaslos:
Addition to the config:
```
Htpasswd: ".htpasswd"
BasicRealm: "test"
```
If Htpasswd file present, `401` error on authentication failure, `200` else.


arnecls:
Code looks good to me.
Please add "github.com/abbot/go-http-auth" to the vendor folder using [gvt](https://github.com/FiloSottile/gvt) and add the changes.

glaslos:
I should add that the basic auth check is rather expensive. With producer set to `Null`, this has a 10x overhead. There are for sure cheaper auth methods for HTTP (cookie?). Might be also worth to investigate HTTP2 which should add significant performance.

arnecls:
Performance wise I've got no problem with this as the feature. It is optional and if someone decides to use an authenticated HTTP endpoint (and/or HTTPS), performance is probably not the first priority anyway.
In a high performance usecase a local UDS connection (consumer.Socket) with a buffered transfer to a secured, centralized backend service is probably a better idea.

arnecls:
All approved: /merged
Thank you! :+1:

PR #65, opened 2017-01-28
Fix format.ProcessJSONs flatten directive

relud:
without this format.ProcessJSON would fail to flatten keys containing json objects, because json objects must be cast to `map[string]interface{}` instead of `map[interface{}]interface{}`


arnecls:
:+1: /merged

Issue #64, opened 2017-01-19
Missing licenses in vendor directory


glaslos:
The package source code in the vendor directory is missing the license information. This might be related to https://github.com/FiloSottile/gvt/issues/46 or https://github.com/FiloSottile/gvt/issues/23
This is probably violating most of the licenses used by packages in the vendor directory.

arnecls:
Thanks for noticing ... let's see how we can fix that quickly ... 🤔

glaslos:
Actually it might be the `.gitignore`, adding `!LICENSE` (+all the other variants) could be the solution.

arnecls:
Interestingly https://github.com/trivago/gollum/tree/master/vendor/github.com/eapache/queue contains a license file

arnecls:
I've forwarded this to legal because I am not 100% sure about the situation here.
If the source of the problem is GVT we might need to switch the vendor tool, because I don't want to adde license files manually each time a new dependency arrives ...

arnecls:
License files should be included for all vendors on the 0.4.5dev branch.

arnecls:
A little update on this issue (details are still being evaluated).
1. We will add all missing license files for the next release (0.4.5)
2. If _requested_ we will also fix this for already released versions*
3. You can always restore the license files by using `gvt restore`

*because licenses can be restored easily and fixing old releases is quite a bit of work.

arnecls:
0.4.5 has been merged, missing license files have been added.
I also did some clarifications with our legal department and (as of now) it should be sufficient to have the License files in the vendor folder.

PR #66, opened 2017-02-14
replacing net.websocket with gorilla.websocket

glaslos:
The old websocket producer was broken. In the godoc of net.websocket they recommend using gorilla.websocket: https://godoc.org/golang.org/x/net/websocket
I replace the old websocket handler with the gorilla one.
I also added an exclude in the .gitignore to add the license file, see #64


arnecls:
That also sneaked in the license vendor rule ^^
Looks ok to me, we should just check if the golang.org/x/net/websocket can be removed now, too.

glaslos:
I removed `net.websocket` from vendor.

arnecls:
👍 I'll merge this once andy had a look at it too

PR #67, opened 2017-02-17
websocket: Adding IgnoreOrigin and minor bug fix

glaslos:
Run into an issue with the websocket server Origin default check. So added a flag to disable it. Also fixed an issue in case a connection upgrade fails.

### DIFF
--- a/docs/producers/websocket.rst
+++ b/docs/producers/websocket.rst
@@ -75,6 +75,10 @@ Parameters
   ReadTimeoutSec specifies the maximum duration in seconds before timing out read of the request.
   By default this is set to 3 seconds.
 
+**IgnoreOrigin**
+  IgnoreOrigin disables the Origin header check and allows connections from any remote host.
+  By default this is set to false.
+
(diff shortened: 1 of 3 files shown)


arnecls:
@andygrunwald have a second look please.
Origin flag reference: https://godoc.org/github.com/gorilla/websocket#hdr-Origin_Considerations
I'm fine with handling it this way.

docs/producers/websocket.rst
```diff
@@ -97,3 +101,4 @@ Example
 	    Address: ":81"
 	    Path:    "/"
 	    ReadTimeoutSec: 3
+        IgnoreOrigin: false
```
andygrunwald: It looks that the indenting does not fit. Don't know (yet) if this will end in a different rendering result.

andygrunwald:
@arnecls Go for it.

glaslos:
Yes, my bad, messed up the formatting in the Readme example

PR #68, opened 2017-03-08
Updated aws kinesis config documentation

gumuz:
Looking at the code, i think these args should be lowercase, right?

(also fixed a typo newset -> newest)

### DIFF
--- a/consumer/kinesis.go
+++ b/consumer/kinesis.go
@@ -84,7 +84,7 @@ const (
 // By default this is set to none.
 //
 // DefaultOffset defines the message index to start reading from.
-// Valid values are either "Newset", "Oldest", or a number.
+// Valid values are either "Newest", "Oldest", or a number.
 // The default value is "Newest".
 //
 // OffsetFile defines a file to store the current offset per shard.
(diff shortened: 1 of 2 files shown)


arnecls:
No, the casing is removed by strings.ToLower, see https://github.com/trivago/gollum/blob/04b1a647f21a6521d6db21b89c6e4278a4ee33ef/consumer/kinesis.go#L180

BTW - this RST docs for plugins get auto generated from the go docs by an internal tool so it is better to fix the godoc documentation and not the RSTs - at least for plugins.
We will open source that generator some time, too, but it is *very*, *very*, _VERY_ hacky at the moment.

gumuz:
I see! I will correct the typo in the kinesis.go file then. 

So the convention is to capitalize the parameters?

arnecls:
In 0.4.x and older there is a - let's call it - "guideline" that parameters ignore case while keys are case sensitive. That means keys were ALWAYS case sensitive but most parameters were case insensitive. In the documentation we tended to capitalize/camel case parameters but in the end it should not matter.

In 0.5.x keys and values (by default) will be case insensitive.

PR #72, opened 2017-03-15
fix typos

mhils:
Just spotted those when not using autocompletion. :-)


andygrunwald:
This was a no brainer. Thank you.

PR #77, opened 2017-03-27
Add Go 1.7 and 1.8 to TravisCI

andygrunwald:


arnecls:
This is already fixed as part of Gollum 0.4.5 which is going to be released next week.

Issue #82, opened 2017-04-03
Add a "key" field to messages


arnecls:
Currently we have a lot of cases were messages are separated into key and value (e.g. after reading from kafka) or where a key has to be generated from a payload (routing, redis, kafka, etc.).
Along with that comes the need to modify keys separately that lead to different approaches and config + code bloat.
By adding keys to messages we will reduce processing overhead and can built much more streamlined pipelines that are easier to understand and debug.

**See/discuss changes in comments**

- Add a "key" to the Message struct ([]byte)
- Consumers reading from key/value stores should set the key
- Producers requiring a key should take the key field into account (can be empty)
- Modulators for keys must be possible (ideally by defining an ApplyTo field or sth. similar)
- Implement a "Copy" formatter that allows copying data between key and value fields

arnecls:
After an internal discussion we agreed on changing this from a dedicated field to a metadata collection. Changed requirements:

- Add a "metadata" field to the message struct (map[string][]byte)
- Consumers can decided where to put the key by config value (default "key")
- Producers can read a required key from metadata by config value
- Formatters may store snippets in metadata fields
- Formatters may read/write metadata fields by config value (default: read/write payload)
- Filters may read from metadata fields
- Routers may react on metadata fields

msiebeneicher:
This task is done for now. Open features will handles by new subtasks.

Issue #81, opened 2017-04-03
Add support for originals in message


arnecls:
The intention of this task is to ease development for plugin writers regarding spooling support and adds more consistency to the way messages are modulated and forwarded.
By removing the support for message modification from the routers we can define an "original" to be generated by the consumer and can be sure no additional modification has been made before the message reaches a producer. By dropping only original messages we avoid double-formatting of messages e.g. sent to a spooling producer.

- The message generated from by a consumer AFTER applying modulators is seen as "original"
- A function "CloneOriginal" has to be implemented in the Message struct
- Remove formatter support from routers (see #80)
- When dropping a message a clone of the original is ALWAYS dropped (with streams set correctly)
- Modulating of messages has to move out of the concrete producer implementation unless SimpleProducer is used as a base struct (i.e. as of the current state BufferedProducer calls Modulate)
- Storing / generating originals has to be removed from all producers doing such a thing

arnecls:
@msiebeneicher from our discussion:
´´´go
type Message struct {
	data			MessageData
	orig			MessageData
	prevStreamID	MessageStreamID // history
	source			MessageSource   // back channel	  <- optional
	timestamp		time.Time		// time of creation  
	sequence		uint64			// weak ordering     <- raus?
}

type MessageData struct {
	key			[]byte			// later
	payload     []byte
	streamID    MessageStreamID
}
```

PR #91, opened 2017-04-06
Bugfix/filter in file producer do not work

msiebeneicher:
* Added integration test case to cover current issue
* Fixed issue that producers did not executed modulators correctly

### DIFF
--- a/core/writerassembly.go
+++ b/core/writerassembly.go
@@ -91,7 +91,6 @@ func (asm *WriterAssembly) Write(messages []*Message) {
 	contentLen := 0
 	for _, msg := range messages {
 		msgCopy := msg.Clone()
-		asm.modulator.Modulate(msgCopy)
 
 		if contentLen+len(msgCopy.Data()) > len(asm.buffer) {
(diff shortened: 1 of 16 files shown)


core/writerassembly.go
```diff
@@ -91,7 +91,6 @@ func (asm *WriterAssembly) Write(messages []*Message) {
 	contentLen := 0
 	for _, msg := range messages {
 		msgCopy := msg.Clone()
```
arnecls: Just for reference (we already talked about this): Until #81 has been implemented drop is broken here as messages does not contain the "originals" any more.

producer/kafka.go
```diff
@@ -514,11 +512,9 @@ func (prod *Kafka) produceMessage(msg *core.Message) {
 
 	if prod.keyFirst {
 		keyMsg := originalMsg.Clone()
-		prod.Modulate(keyMsg)
```
arnecls: Another breaking change that will be fixed once #82 has been implemented

Issue #80, opened 2017-04-03
Rename "stream" plugins to "router"


arnecls:
Currently there is a lot of confusion in the wording between the "stream" concept and the "stream" plugins. As of this we want to rename all "stream" plugins to "router".

msiebeneicher:
First PR for renaming: https://github.com/trivago/gollum/pull/92

PR #102, opened 2017-04-26
Feature/add support for original data in messages

msiebeneicher:
Pull request for issue #81 

- A function "CloneOriginal" has to be implemented in the Message struct
- Added MessageData struct to Message struct instead of direct data []byte
- Added new method CloneOriginal() to *Message
- Added Filter and Formatter interfaces
- Updated all existing core filter and formatter
- Added FilterModulator and FormatterModulator wrapper
- Added FilterArray and FormatterArray types beside exiting ModulatorArray
- Renamed modulator related go files (sub-package creation) 
- New message now get consumer stream as initial stream setting
- Removed formatter support from routers (see #80)
- Removed deprecated ModulateResultRoute and ModulateResultHandled status

- Increased code coverage


core/pluginconfigreaderwitherror.go
```diff
@@ -236,29 +236,82 @@ func (reader PluginConfigReaderWithError) GetPluginArray(key string, defaultValu
...
+	}
+	if formatters == nil {
+		return defaultValue, nil
+	}
+
+	return formatters, nil
+}
+
+// getModulatorArrays returns ModulatorArray, FilterArray and FormatterArray arrays
+func (reader PluginConfigReaderWithError) getModulatorArrays(key string, logScope tlog.LogScope) (ModulatorArray, FilterArray, FormatterArray, error) {
```
msiebeneicher: @arnecls : please check the implementation here

core/modulator_filter.go
```diff
@@ -0,0 +1,103 @@
...
+	return filterModulator.Filter.ApplyFilter(msg)
+}
+
+// drop set filter DropStreamID to message and returns ModulateResult
+func (filterModulator *FilterModulator) drop(msg *Message) ModulateResult {
+	// if not drop stream => discard | else drop
+	filterDropStreamID := filterModulator.Filter.GetDropStreamID()
+	if filterDropStreamID != InvalidStreamID {
+		msg.SetStreamID(filterDropStreamID)
+		return ModulateResultDrop
```
arnecls: Should be ModulateResultRoute

core/pluginconfigreaderwitherror.go
arnecls: I would rather split this in 3 separate functions.
1. Clearer, shorter code + less return variables
2. Stricter typing / error handling (filter arrays must not contain formatters)
3. Code duplication is so minor it can be neglected

core/simplerouter.go
```diff
@@ -20,6 +20,16 @@ import (
...
+//    Type: "router.Broadcast"
+//    Stream: "foo"
+//    TimeoutMs: 200
+//    Filters:
+//	- filter.RegExp:
+// 	  Expression: "[a-zA-Z]+"
+//
 type SimpleRouter struct {
 	id         string
 	modulators ModulatorArray
```
arnecls: This should be a filter array

core/simplerouter.go
```diff
@@ -35,7 +45,13 @@ func (router *SimpleRouter) Configure(conf PluginConfigReader) error {
 	router.Log = conf.GetLogScope()
 	router.Timeout = nil
 	router.streamID = conf.GetStreamID("Stream", GetStreamID(conf.GetID()))
-	router.modulators = conf.GetModulatorArray("Modulators", router.Log, ModulatorArray{})
+
+	// init router.modulators with filters only
+	filters := conf.GetFilterArray("Filters", router.Log, FilterArray{})
+	for _, filter := range filters {
+		filterModulator := NewFilterModulator(filter)
```
arnecls: You can spare yourself this code if you have a filter array instead of a modulator array

msiebeneicher: If i use the FilterArray i need to implement the Modulate() method again which i want to avoid.  I think it's easier here to handle only the logic in the router to use filters only. This is also more flexible in case we get other modulator types later on.

core/simplerouter.go
msiebeneicher: see comment below

core/modulator_filter.go
msiebeneicher: This suggestion do not fit the current implementation of the simple-consumers and -producers. see: 

* https://github.com/trivago/gollum/blob/feature/add_support_for_original_data_in_messages/core/simpleconsumer.go#L230
* https://github.com/trivago/gollum/blob/feature/add_support_for_original_data_in_messages/core/simpleproducer.go#L238

core/pluginconfigreaderwitherror.go
msiebeneicher: updated

core/simplerouter.go
arnecls: Summary of our internal discussion:
The Filter interface is not recursive. This happens because the return value leaves it to the caller on who to react on a "reject". This means the caller needs to get the DropStreamID from the Filter. An array cannot easily implement GetDropStreamID as in case of a reject the DropStreamID is bound to a specific index for that specific call.
We have several options to fix this:
1. Let filter return the drop stream id. This puts additional burden on the maintainer of the filter (return the correct config value in case of reject)
2. Write a GetDropStreamID for FilterArray that re-evaluates alle filters. This collides with filters that store internal counters like the rate limiter.

As of this we decided to leave it as is and do the "enrichment" on the next higher level, i.e. the modulator. This is currently the easiest implementation but is still bad in terms of type safety and readability.

I.e.: This is an architectural TBD and might change with a later PR

arnecls:
:beer: yolo

Issue #88, opened 2017-04-03
Add an optional http health endpoint


arnecls:
Currently we (and probably others, too) use the metrics provided by gollum to do health checking.
A better way would be to allow plugins to return a health state that can be queried by a given http endpoint.

- Add a new commandline flag to gollum to open up an http endpoint like `gollum --health=:80`
- A general health check should be added to `/health` (maybe based on error/warning messages per second)
- Health checks should be added to `/health/plugin_id`
- Plugins can implement a HealthCheck interface `GetHealth() (Status, string)` where status can be of ok, warning or critical
- We should discuss wether to use different http status codes and if yes, which ones depending on the returned status

ppar:
Fixed by PR https://github.com/trivago/gollum/pull/105 commit  f1a0109ef6b5ea18bb2225e5840c92b0dba76ece

PR #105, opened 2017-04-28
Health check endpoint, with sample implementation for prod.HTTPRequest (#88)

ppar:
This PR implements an HTTP health check endpoint in Gollum where plugins can add their own health check hooks and adds a couple of simple health checks in the `producer.HTTPRequest` plugin.

~~Classes~~Structs ~~inheriting from~~ embedding `SimpleConsumer`, `SimpleProducer` or `SimpleRouter` can simply call `.AddHealthCheck(callback)` to insert a health check. `callback`'s signature must match `tgo.thealthcheck.CallbackFunc`: `func()(code int, body string)`

The health checks offered look like this:

```
$ curl http://localhost:8012/
/_ALL_
/HttpOut01/lastError
/_PING_
/HttpOut01/SimpleProducer/pluginstate
/HttpOut01
```

Health checks can be queried individually by path or all in sequence using `/_ALL`:

```
$ curl http://localhost:8012/_ALL_
/_PING_ 200 PONG
/HttpOut01/SimpleProducer/pluginstate 200 ACTIVE: Active
/HttpOut01 200 Done

/HttpOut01/lastError 200 OK
```

Notes:

- ~~The `healthcheck` package is independent from Gollum and could be moved to the `tgo` repository~~
- ~~Command line parameter for the port is still missing in main.go~~
- Duplicate "\n"s should be cleaned up
- Health checks could be en-/disabled (and maybe fine-tuned) in plugin configuration
- What constitutes a proper health check in each plugin is still an open question. Should we measure past errors, or actively probe the service we're talking to? Can we do the latter idempotently? E.g. reading from an S3 bucket or GETing a web service do not guarantee we can also write or POST to them, but the latter are rarely safe (or free) operations either.


core/plugin.go
```diff
@@ -85,6 +85,16 @@ var (
 		metricPluginsStopping,
 		metricPluginsDead,
 	}
+
+	// Has to be index parallel to PluginState*
+	stateToDescription = []string{
```
arnecls: It might be a better idea to solve this translations with struct tags (https://golang.org/pkg/reflect/#StructTag).
Programatically more complex to resolve but avoids having to change two places in a file where the second part may be forgotten.

healthcheck/healthcheck.go
```diff
@@ -0,0 +1,216 @@
...
+// probes are executed by issuing requests to the endpoints' named
+// URL paths.
+//
+// GETing the "/" path provides a list of registered endpoints, one
+// per line. GETing "/_ALL_" probes each registered endpoint sequentially,
+// returning each endpoint's path, HTTP status code and body per line.
+//
+// The package works as a "singleton" with just one server in order to
+// avoid cluttering the main program by passing handles around.
+package healthcheck
```
arnecls: A package just for one file seems a bit overblown for me.
Sure, core is overloaded at the moment but the package should at least be core/healthcheck.
If it is totally generic it should go to github.com/trivago/tgo

core/simpleproducer.go
```diff
@@ -123,6 +125,19 @@ func (prod *SimpleProducer) Configure(conf PluginConfigReader) error {
 	dropStreamID := StreamRegistry.GetStreamID(conf.GetString("DropToStream", DroppedStream))
 	prod.dropStream = StreamRegistry.GetRouterOrFallback(dropStreamID)
 
+	// Simple health check for the plugin state
+	//   Path: "/<plugin_class>/<plugin_id>/SimpleProducer/pluginstate"
```
arnecls: As pluginID has to be unique over all plugin types we don't really need the plugin_class.
So it should be /<plugin_id>.
Simple producer may include a basic health check like this, but I would prefer function chaining, i.e. creating a list of functions to call as a health check and combining their results (TBD)

main.go
```diff
@@ -115,6 +116,15 @@ func main() {
 		defer server.Stop()
 	}
 
+	// Health Check endpoint
+
+	if true {
```
arnecls: leftover -> remove

main.go
```diff
@@ -115,6 +116,15 @@ func main() {
...
+		// TODO: package, config
+		healthcheck.Configure(":8012")
```
arnecls: Should be defined by command line flags (see flags.go)

producer/httprequest.go
```diff
@@ -78,9 +82,69 @@ func (prod *HTTPRequest) Configure(conf core.PluginConfigReader) error {
 	prod.encoding = conf.GetString("Encoding", "text/plain; charset=utf-8")
 	prod.rawPackets = conf.GetBool("RawData", true)
 
+	//prod.Log.Debug.Printf(
+	//"protocol=%s, host=%s, port=%s, address=%s, encoding=%s, rawPackets=%d",
+	//	prod.protocol, prod.host, prod.port, prod.address, prod.encoding, prod.rawPackets)
+
+	// Health check to ping the backend with an HTTP GET
+	healthcheck.AddEndpointPathArray(
```
arnecls: Again, this seems to much effort for this.
Should be something like `prod.AddHealthCheck(prod.CheckHealth)`

healthcheck/healthcheck.go
```diff
@@ -0,0 +1,216 @@
...
+//  listenAddr: an address understood by http.ListenAndServe(), e.g. ":8008"
+func Configure(listenAddr string) {
+	// Create the HTTP server
+	server = &http.Server{
+		Addr:    listenAddr,
+		Handler: serveMux,
+	}
+
+	// Add default wildcard handler
+	serveMux.HandleFunc("/",
```
arnecls: Have a look at this: https://github.com/gorilla/mux
This might drastically simplify this.

core/simpleproducer.go
ppar: I think exposing each health check behind its own URL path gives more granularity for the user; they can e.g. give different priorities to different checks or choose to ignore some. OTOH, the /_ALL path still lets you run all checks at once. 

Skipping /<plugin_class> makes sense, although it could be informative to show also what type of plugin each check belongs to.

healthcheck/healthcheck.go
ppar: It is completely generic like the description says; the reason it's included in this PR is to have the whole change visible in one place.

main.go
ppar: Placeholder for flag parsing, like the description says

producer/httprequest.go
ppar: We'd have to implement an `.AddHealthCheck()` in each of the `core.Simple***` classes separately since the plugins don't have a common ancestor, and with the proposed syntax, add some reflection to append the function name to the endpoint's path, but it could save a couple of lines per endpoint.

healthcheck/healthcheck.go
ppar: We use it in another Go project but I don't see a lot of benefit here -- there are only a couple of statically defined paths, the rest are created with `.AddEndpoint()`

core/plugin.go
ppar: I don't see how struct tags could work here; we need a mapping from enum values to strings, not string values for a struct's fields. It's analogous to the stateToMetric mapping above for the same enums.

arnecls: Hence the "might" - I wasn't quite sure if these tags are available for other types, too, but apparently they are not 🙄  In that case - plain old mapping it is.

healthcheck/healthcheck.go
arnecls: Good point, but if you include a dependency update this will show up here, too, as we check in the vendor folder, too. So I would still recommend to put this one into tgo so that we can use it for other projects.
... OR we create a completely separate package ... maybe something to discuss on Tuesday.

core/simpleproducer.go
arnecls: As we already discussed in the office - the question is we really need multiple health checks in a plugin or not.

main.go
arnecls: then make it a "modify" 😀

producer/httprequest.go
arnecls: Also as discussed in the office.
My original idea was to implement the health check as an interface. That way it would be sufficient to implement one method and registration happens during startup.
Good:
- easy for the developer
Bad:
- Multiple endpoints would be a breaking change
- "magic" vs. type hierarchy ... although it's debatable what's really worse

Using the type hierarchy is possible as currently all plugins derive directly or indirectly from Simple*. One can choose not to but in that case you're going all in anyway. Aka - if you want all the pain, you have to deal with it.
But even here a `RegisterEndpoint(name string, callback func()result)` would work fine.
Good:
- Multiple endpoints are possible by default
Bad:
- "magic" vs. type hierarchy (see above, not sure if good or bad)

I would say, let's make a pair session on Tuesday and see how both solutions feel and select by that.

ppar: IMO making as little assumptions about plugins as possible ( == implicitly allowing multiple endpoints here) is more valuable than providing shorter syntax. 

From a new developer's perspective, diving through the initialisation code to find out how & where the endpoints are defined via the interface trick could also be more time-consuming than just looking up the single row where the plugin calls tgo.healthcheck.AddEndpoint(). 

And the healthcheck.AddEndpoint() approach would let us easily add health checks anywhere in the code, e.g. somewhere in the core classes if such a need comes up in the future.

But I agree, let's look at the details & relative merits IRL.

ppar:
Changes applied as discussed

core/simpleproducer.go
```diff
@@ -123,9 +125,28 @@ func (prod *SimpleProducer) Configure(conf PluginConfigReader) error {
 	dropStreamID := StreamRegistry.GetStreamID(conf.GetString("DropToStream", DroppedStream))
 	prod.dropStream = StreamRegistry.GetRouterOrFallback(dropStreamID)
 
+	// Simple health check for the plugin state
+	//   Path: "/<plugin_id>/SimpleProducer/pluginstate"
+	prod.AddHealthCheckAt("/SimpleProducer/pluginState", func()(code int, body string) {
```
arnecls: Remove SimpleProducer from the path because this creates a semantic link to the actual implementation which might get problematic in the future.

main.go
```diff
@@ -101,20 +102,34 @@ func main() {
...
+			fmt.Printf("%s", err)
+			return
+		}
+		thealthcheck.Configure(address)
+
+		go thealthcheck.Start()
+		defer thealthcheck.Stop()
+
+		// Add a static "ping" endpoint
+		thealthcheck.AddEndpoint("/ping", func()(code int, body string){
```
arnecls: If we stick to "reserved terms are of the format \_*\_" this should be /\_PING\_ ... just a thought

main.go
```diff
@@ -160,6 +175,19 @@ func main() {
 	coordinater.Run()
 }
 
+func parseAddress(address string) (string, error) {
```
arnecls: have a look at tgo/tnet/SplitAddress

producer/httprequest.go
```diff
@@ -78,9 +82,60 @@ func (prod *HTTPRequest) Configure(conf core.PluginConfigReader) error {
 	prod.encoding = conf.GetString("Encoding", "text/plain; charset=utf-8")
 	prod.rawPackets = conf.GetBool("RawData", true)
 
+	// Default health check to ping the backend with an HTTP GET
+	prod.AddHealthCheck(func()(int, string){
+		return prod.healthcheckPingBackend()
+	})
```
arnecls: You don't need the func() wrapper here

producer/httprequest.go
```diff
@@ -78,9 +82,60 @@ func (prod *HTTPRequest) Configure(conf core.PluginConfigReader) error {
...
+
+
 	return conf.Errors.OrNil()
 }
 
+func (prod *HTTPRequest) healthcheckPingBackend() (int, string) {
+	code, body, err := httpRequestWrapper(http.Get(prod.address))
+	if err != nil {
+		return code, fmt.Sprintf("%s", err)
```
arnecls: err.Error() returns a string

producer/httprequest.go
```diff
@@ -78,9 +82,60 @@ func (prod *HTTPRequest) Configure(conf core.PluginConfigReader) error {
...
+	if err != nil {
+		// Fail
+		return resp.StatusCode, "", err
+	}
+
+	respBodyString := fmt.Sprintf("%s", respBody)
+
+	err = nil
+	if resp.StatusCode != http.StatusOK {
+		err = errors.New(fmt.Sprintf("%d %s", resp.StatusCode, respBodyString))
```
arnecls: fmt.Errorf()

main.go
ppar: Actually we should use https://golang.org/pkg/net/#SplitHostPort since SplitAddress allows for a the `proto:` prefix that we're not using in its input. But neither accepts a plain port number, without the preceding `:`.

ppar:
Changes done, should be ready for merging

arnecls:
:+1:

PR #110, opened 2017-05-04
Make Filter a "recursive" interface again

arnecls:
This change allows to store the DropStreamID with a filter result.
As we only have a "yes" or "no" result we can define "yes" as 0 and "no" as anything else, which means it can be a stream id, too. As it is close (or even completely?) impossible to generate a 0 result with fnv1a on the standard ASCII character set there either is or should not be any collision with 0.


core/filterresult.go
```diff
@@ -0,0 +1,36 @@
...
+	return FilterResult(stream)
+}
+
+// GetStreamID converts a FilterResult to a stream id
+func (f FilterResult) GetStreamID() MessageStreamID {
+	if f == FilterResultMessageAccept {
+		return InvalidStreamID
+	}
+	return MessageStreamID(f)
+}
```
msiebeneicher: because the new file names like `modulator_filter.go` i would continue with that and use `modulator_filter_result.go` for this file.

arnecls: That's something that we already talked about.
If you look at the go std lib they do never use underscore in filenames. That is because you can add additional information to a file by that. Like _test, _linux, _x86 and so on.
As of this I consider modulator_* an intermediate solution.

I already started with a branch to separate core into more subpackages which is IMO the proper way to go here.

Issue #119, opened 2017-05-11
Create a new formatter for setting meta data


msiebeneicher:
This issue is related to PR #117 and a subtask from #82 

We need a new formatter which is be able to copy/set meta data to a message.

PR #120, opened 2017-05-11
Added MetaDataCopy formatter plugin

msiebeneicher:
PR for issue #119

### DIFF
--- a/core/message.go
+++ b/core/message.go
@@ -73,12 +73,12 @@ func (meta MetaData) SetValue(key string, value []byte) {
 }
 
 // GetValue returns a meta data value by key
-func (meta MetaData) GetValue(key string, defaultValue []byte) []byte {
+func (meta MetaData) GetValue(key string) []byte {
 	if value, isSet := meta[key]; isSet {
 		return value
 	}
 
-	return defaultValue
+	return []byte{}
 }
 
 // ResetValue delete a meta data value by key
--- a/core/messagehelper.go
+++ b/core/messagehelper.go
@@ -2,8 +2,9 @@ package core
 
 import (
 	"github.com/trivago/tgo/tlog"
-	"strings"
 )
+
+const APPLY_TO_PAYLOAD = "payload"
 
 // GetAppliedContent is a func() to get message content from payload or meta data
 // for later handling by plugins
@@ -11,11 +12,9 @@ type GetAppliedContent func(msg *Message) []byte
 
 // GetAppliedContentFunction returns a GetAppliedContent function
 func GetAppliedContentFunction(applyTo string) GetAppliedContent {
-	parts := strings.Split(applyTo, ":")
-
-	if parts[0] == "meta" {
+	if applyTo != "" && applyTo != APPLY_TO_PAYLOAD {
 		return func(msg *Message) []byte {
-			return msg.MetaData().GetValue(parts[1])
+			return msg.MetaData().GetValue(applyTo)
 		}
 	}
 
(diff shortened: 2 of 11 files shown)


core/message.go
```diff
@@ -73,12 +73,12 @@ func (meta MetaData) SetValue(key string, value []byte) {
...
 
 // GetValue returns a meta data value by key
-func (meta MetaData) GetValue(key string, defaultValue []byte) []byte {
+func (meta MetaData) GetValue(key string) []byte {
 	if value, isSet := meta[key]; isSet {
 		return value
 	}
 
-	return defaultValue
+	return []byte{}
```
arnecls: TBD
Maybe returning nil is a better idea here?
The thing is - you cannot distinguish between an empty value and a non-existing value here.
On the other hand if you always return a "valid" result you don't need to check for errors ...

core/messagehelper.go
```diff
@@ -15,7 +15,7 @@ func GetAppliedContentFunction(applyTo string) GetAppliedContent {
 
 	if parts[0] == "meta" {
```
arnecls: TBD
The "meta" keyword does not add any information to the argument. We only use it so that a metadata key called "payload" is allowed, too. Without that we could just say "anything that is NOT payload is treated as a metadata key".
... if you continue thinking this way, "payload" is just a special metadata field ... interesting.
That makes the payload field an access-time optimisation.

format/metadatacopy.go
```diff
@@ -0,0 +1,109 @@
...
+
+// MetaDataCopy formatter plugin
+// Copy data from defined payload or meta data to set meta data field
+//
+// Configuration example
+//
+// - format.MetaDataCopy:
+//   WriteTo:
+//     - hostname: 	# meta data key
+//       Modulators:	# further modulators
```
arnecls: Do we really need the Modulators key here?
Looks superflous to me

format/metadatacopy.go
```diff
@@ -0,0 +1,109 @@
...
+	writeToConfig := core.NewPluginConfig("", "format.MetaDataCopy.WriteTo")
+
+	for _, v := range metaData {
+		if converted, isMap := v.(tcontainer.MarshalMap); isMap {
+			for keyMetaData, valueMetaData := range converted {
+
+				writeToConfig.Read(valueMetaData.(tcontainer.MarshalMap))
+				reader := core.NewPluginConfigReaderWithError(&writeToConfig)
+
+				modulator, err := reader.GetModulatorArray("Modulators", format.Log, core.ModulatorArray{})
```
arnecls: See remark above.
Also you can access keys by path semantics, e.g. `keyMetaData+"/Modulators"`

format/metadatacopy.go
msiebeneicher: superfluous? Thats the most powerfully feature in this pr! ;)

core/message.go
arnecls: We decided to keep it as is. Reasons:
- Currently we don't need to check for existence
- We DO need to check for empty values anyways
- If we every need to check for existence a GetValueOpt function could easily be implemented

core/messagehelper.go
```diff
@@ -2,20 +2,19 @@ package core
 
 import (
 	"github.com/trivago/tgo/tlog"
-	"strings"
 )
 
+const APPLY_TO_PAYLOAD = "payload"
```
arnecls: Constants should follow the golang formatting rules:
http://stackoverflow.com/questions/22688906/go-naming-conventions-for-const

Issue #99, opened 2017-04-24
Rename "Drop*" to "Retry*"


arnecls:
The term Drop lead to a lot of confusion.
Most people thing of "drop" as an destructive operation (like in mysql, ip, etc.). As of this drop was mistaken as a discard in the past quite often.
The "Drop" mechanism was originally thought of as kind of an exception handling mechanism: If you cannot deliver a message, drop it, so somebody else might catch it. So basically it implements a "retry by routing" operation.

- Find all occurrences of "Drop*" and replace them by "Retry*"
- Make it clear that this operation is a "retry in case of error" in the documentation

Issue #86, opened 2017-04-03
Stream based metrics


arnecls:
Some producers implement metrics based on their input streams.
Production showed that this is something required for all streams to have a better overview and to be able to debug problems.

Metrics should be implemented on router level.

PR #122, opened 2017-05-15
Rename Drop* to Fallback*

arnecls:
Implementing issue #99
I decided on using "Fallback" instead of Retry as it is shows the intend more clearly.
Filters are a special case as "fallback" is not a good match here. As of this we have a special naming here.

- FallbackStream is used instead of DropStream for producers
- FilteredStream is used instead of DropStream for filters
- DropMessage renamed to RouteOriginal
- Adjusted all other occurrences to match the new naming pattern
- The \_DROPPED\_ stream has been removed, by default messages are discarded
- Drop stream metrics have been removed as #86 will make this metric obsolete


core/messagehelper_test.go
```diff
@@ -79,7 +79,7 @@ func TestDropMessage(t *testing.T) {
 
 	msg := NewMessage(nil, []byte("foo"), 0, mockRouter.StreamID())
 
-	err := DropMessage(msg)
+	err := RouteOriginal(msg, msg.GetRouter())
```
msiebeneicher: This unit test should move to the RouteOriginal() class tests

core/messagehelper_test.go
```diff
@@ -111,7 +111,7 @@ func TestDropMessageByRouter(t *testing.T) {
 	// create message and test
 	msg := NewMessage(nil, []byte("foo"), 0, mockRouterA.StreamID())
 
-	err := DropMessageByRouter(msg, &mockRouterB)
+	err := RouteOriginal(msg, &mockRouterB)
```
msiebeneicher: same here

core/bufferedproducer.go
```diff
@@ -108,30 +108,32 @@ func (prod *BufferedProducer) Enqueue(msg *Message, timeout *time.Duration) {
...
 			prod.Log.Error.Print("Producer: ", prod.id, "State: ", prod.GetState(), ", Router: ", StreamRegistry.GetStreamName(msg.StreamID()))
-			prod.Drop(msg)
+			prod.TryFallback(msg)
 		}
 	}()
 
 	// Run modulators and drop message if necessary
 	result := prod.Modulate(msg)
 	switch result {
-	case ModulateResultDrop:
```
msiebeneicher: no ModulateResultFallback handling instead?

arnecls: But it's there ..? It just moved below discard

msiebeneicher: sry - did not saw it correctly in the unified view - ignore the comment

Issue #115, opened 2017-05-09
Remove sequence number


arnecls:
Gollum adds a sequence number to each message.
Originally the idea was to give messages a kind of "unique" id.
This however does not work because of multiple reasons.

- Sequence numbers are bound to a consumer (i.e. there are collisions between consumers)
- Sequence numbers are reseted after a restart

This means that sequence number cannot be used for anything meaningful.
What it is used for currently is for hashing and loadbalancing. This two usecases can easily be implemented in the router / formatter itself.

- Remove sequence number from messages and all functions associated with it
- Where used replace it with a locally stored number

PR #123, opened 2017-05-15
Remove sequence number from message

arnecls:
Implementing issue #115
Messages don't store a sequence number anymore.
Any plugins relying on sequence numbers need to track one by themselves.

The message protobuf format contains a breaking change as the sequence number filed was labeled as required. As of this reading protobufs from 0.4.x will probably fail.


core/message.proto
```diff
@@ -5,7 +5,7 @@ message serializedMessage {
         required uint64 StreamID = 1;
         required uint64 PrevStreamID = 2;
         required int64 Timestamp = 3;
-        required uint64 Sequence = 4;
+        //required uint64 Sequence = 4;
```
msiebeneicher: remove the line

arnecls: I'll suggest we leave that one here because this is the breaking change mentioned in my initial comment. We might still require it in the future.

Issue #118, opened 2017-05-11
Update the modulators for meta data usage


msiebeneicher:
This issue is related to PR #117 and a subtask from #82 

Goal is to update all modulator for optional meta data usage:
`ApplyTo: "payload" # [payload, meta:key]`

PR #124, opened 2017-05-16
Feature/update modulators for meta data usage

msiebeneicher:
This PR is related to #118 

- Added `applyTo` handling to all core formatters
- Updated formatter docs


format/hostname.go
```diff
@@ -52,19 +52,31 @@ func (format *Hostname) Configure(conf core.PluginConfigReader) error {
...
 	hostname, err := os.Hostname()
 	if err != nil {
 		format.Log.Error.Print(err)
 		hostname = "unknown host"
 	}
 
-	dataSize := len(hostname) + len(format.separator) + msg.Len()
-	payload := core.MessageDataPool.Get(dataSize)
+	lenSeparator := len(format.separator)
+	if lenSeparator > 0 {
```
arnecls: What if someone wants to add the hostname without a separator?

arnecls: Other formatters follow this pattern, too.
It hink in case of no separator we should do the logical thing - don't use a separator.
Storing just the separator is a totally different thing and can be achieved by using the clear formatter beforehand.

msiebeneicher:
As discussed with @arnecls: “no separator handling” reverted

Issue #85, opened 2017-04-03
Improve 0.5.0 config layout


arnecls:
After an internal discussion we came to the point that the current Gollum 0.5.0 format can be improved further in terms of maintainability and reusability.
As we're pushing a config change with 0.5.0 anyway it does make sense to add this on top of the current changes even though they have not been released yet.

Proposed example:

```yaml
WriteToKafka:
  Type: Producer
  Stream: "*"
  Modulators:
    - filter.Rate
    - format.Hostname
    - format.Envelope
      PostFix: "\r\n"
  Plugins:
    - producer.Kafka:
      Broker: ["192.168.0.1","192.168.0.3","192.168.0.2"]
      Topic: "*"
    - producer.Elasticsearch:
      Master: "192.168.10.1"
      Index: "*"
```

To achieve this we have to adjust the base types and separate the actually plugin configuration from the "standard framework" configuration.
Creating multiple plugins out of a single config block can be done by generating multiple config blocks from one block in the config parsing section. As an example: the config above would be separated into two blocks with a producer.Kafka as single item and a producer.Elasticsearch as single item.

arnecls:
If the preprocessing generates the following configs out of the example above, we don't need to change anything in the handling of the configuration:

```yaml
WriteToKafka0:
  Type: producer.Kafka
  Stream: "*"
  Modulators:
    - filter.Rate
    - format.Hostname
    - format.Envelope
      PostFix: "\r\n"
  Broker: ["192.168.0.1","192.168.0.3","192.168.0.2"]
  Topic: "*"

WriteToKafka1:
  Type: producer.ElasticSearch
  Stream: "*"
  Modulators:
    - filter.Rate
    - format.Hostname
    - format.Envelope
      PostFix: "\r\n"
  Master: "192.168.10.1"
  Index: "*"
```

msiebeneicher:
The feature is ready to merge. Here the example config in the final state:

```yml
WriteToKafka:
  Type: aggregate
  Stream: "*"
  Modulators:
    - filter.Rate
    - format.Hostname
    - format.Envelope
      PostFix: "\r\n"
  Aggregate:
    kafka:
      Type: producer.Kafka
      Broker: ["192.168.0.1","192.168.0.3","192.168.0.2"]
      Topic: "*"
    elasticsearch:
      Type: producer.Elasticsearch
      Master: "192.168.10.1"
      Index: "*"
```

PR #127, opened 2017-05-17
Feature/improve config layout

msiebeneicher:
This PR is related to #85 

- Added config aggregation feature
- Made Config() testable and added unit tests

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -23,7 +23,10 @@ import (
 	"gopkg.in/yaml.v2"
 	"io/ioutil"
 	"reflect"
+	"strings"
 )
+
+const pluginAggregate =  "aggregate"
 
 var (
 	consumerInterface = reflect.TypeOf((*Consumer)(nil)).Elem()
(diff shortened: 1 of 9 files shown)


core/config.go
```diff
@@ -54,30 +57,63 @@ func NewLogScope(conf *PluginConfig) tlog.LogScope {
...
+				subConfig, err := tcontainer.ConvertToMarshalMap(subConfigValues, strings.ToLower)
+				if err != nil {
+					tlog.Error.Print("Error in plugin config ", subPluginsId, err)
+					continue
+				}
+
+				// set up sub-plugin
+				pluginConfig := NewPluginConfig(subPluginsId, "")
+
+				delete(configValues, "Aggregate")
```
arnecls: The type should also be deleted, shouldn't it?
It will probably be overwritten, but if the writer forgets to set it in a sub plugin the sub plugin type will be "aggregate".

msiebeneicher: true - i will delete the type also

Issue #83, opened 2017-04-03
Implement a "BatchedProducer"


arnecls:
A lot of plugins use the batching concept on top of a BufferedProducer.
This should be generalised, especially as a Batch does not require a buffer.

- Have a look at producer.File and similar implementations to extract the pattern into a common base type
- Derive all batch-style producers from this new base type

PR #131, opened 2017-05-19
Feature/batched producer implementation

msiebeneicher:
This PR is related to issue #83:

- Added BatchedProducer implementation
- Updated InfluxDB producer to use BatchedProducer
- Updated Kinesis producer to use BatchedProducer
- fixed profiler consumer output


msiebeneicher:
InfluxDB and Kinesis are already updated to new BatchedProducer implementation.
The following producers need to check if we can update the implementations there also:

- [ ] s3
- [ ] scribe
- [ ] socket
- [ ] spoolfile
- [ ] statsd

Feel free to merge or to update this branch in the next two weeks

PR #133, opened 2017-06-04
Added encodekey option to enable encoding of key

mahdi-hdi:
If the key of kafka message has binary value.Then, it's not possible to separate the key from a message by looking for keySeparator. So, we must encode key before prepending.

It uses the encodeKey option to enable/disable encoding. If encodeKey has been enabled then key will be encoded using base64 and then prepended to message.


arnecls:
This change will become unnecessary with 0.5.x.
You can save the key to a metadata field and apply a standard base64 encode formatter to it.

PR #134, opened 2017-06-05
add an option to slow down consumer.File's polling when reaching EOF

relud:
I'm using `consumer.File` and it's stat'ing a slow moving log file 1100-1800 times a second, which is causing performance issues.


arnecls:
Looks ok to me, but this is a symptomatic fix. The main problem is that there are too many tests.
I created #137 to fix the core issue in 0.5.x.

PR #140, opened 2017-06-09
make producer.Kinesis respect api limits

relud:
The kinesis api has hard limits that require no more than 500 records or 5MB per PutRecords call, and no more than 1MB per record.

This change makes sure that gollum respects those three limits by making multiple PutRecords calls if needed, and making more records if needed.

The result is that BatchMaxMessages and RecordMaxMessages can safely be set to higher values that would have previously caused errors.

### DIFF
--- a/producer/kinesis.go
+++ b/producer/kinesis.go
@@ -115,9 +115,101 @@ const (
 )
 
 type streamData struct {
-	content            *kinesis.PutRecordsInput
-	original           [][]*core.Message
+	content            []*kinesis.PutRecordsInput
+	original           [][][]*core.Message
+	streamName         string
 	lastRecordMessages int
+	lastRequestSize    int
+	recordMaxMessages  int
+	recordMaxSize      int
+	requestMaxSize     int
+	requestMaxRecords  int
+}
+
+func newStreamData(streamName string, recordMaxMessages int) *streamData {
+	return &streamData{
+		content: make([]*kinesis.PutRecordsInput, 0, 1),
+		original: make([][][]*core.Message, 0, 1),
+		streamName: streamName,
+		lastRecordMessages: 0,
+		lastRequestSize: 0,
+		recordMaxMessages: recordMaxMessages,
+		recordMaxSize: 1<<20, // 1 MB per record is the api limit
+		requestMaxSize: 5<<20, // 5 MB per request is the api limit
+		requestMaxRecords: 500, // 500 records per request is the api limit
+	}
+}
+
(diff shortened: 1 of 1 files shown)


arnecls:
If I remember correctly the ratelimits were indirectly enforced by the batch size chosen.
Need to check that.

relud:
it's true that rate limits can be enforced using batch size, but if gollum message size is unpredictable, and RecordMaxMessages is high enough, then some kinesis records may unnecessarily exceed the 1MB record size limit

arnecls:
Ah, ok - that's correct.
The current batch implementation only looks at the count, not the size.

producer/kinesis.go
```diff
@@ -115,9 +115,101 @@ const (
...
+	if len(input.Records) == 0 {
+		return sd.AddRecord()
+	} else {
+		return input.Records[len(input.Records)-1]
+	}
+}
+
+func (sd *streamData) AddMessage(delimiter []byte, data []byte, streamID core.MessageStreamID, msg *core.Message) {
+	record := sd.GetRecord()
+	if len(record.Data) > 0 {
```
arnecls: A benefit from encapsulating the request in an extra type is that you can store the size for each request in there, too. This will make the check for "am I too large" a simple
```go
func (sr streamRecord) OverLimt(data, delimiter []byte) {
    dataSize := len(data)
    if len(sr.records) > 0 {
        dataSize += len(delimiter)
    }
    return sr.currentSize + dataSize > sr.Limit
}

// ...
// func (sd *streamData) AddMessage ... {
    record := sd.GetCurrentRecord()
    if record.OverLimit(data,delimiter) {
        record = sd.AddNewRecord(/* ... */)
    }
    record.AddMessage(/* ... */)
```

This will make it a lot clearer to understand and a lot errorprone as you don't have to expose internal state.
E.g. it was very confusing for me why you check for `record.Data` twice though it wasn't directly visible where it is modified.

relud:
I'm no longer able to work on this

andygrunwald:
@relud Thanks for letting us know. Do you still think it would make sense to integrate into Gollum?
If yes, we might want to check it out in detail.

relud:
yes, I definitely think it makes sense. these limits are hard limits, that aws doesn't allow limit increases for.

andygrunwald:
Okay, thank you. I reopened the PR. I can't promise anything in terms of timeline, but we will check it out if we can make it. Thank you for your effort.

Issue #138, opened 2017-06-07
Configuration by struct tags


arnecls:
Most configure methods are just assignments to members of the plugin struct.
Adding these simple assignment as struct tags will reduce the complexity of most configure methods and provide an inline documentation accessible to e.g. the document generator tool.

PR #143, opened 2017-06-13
Reflection based configure

arnecls:
Implementation of #138.
This PR changes the configuration process to a completely reflection based one.
Plugins will be configured in a two step, recursive process:
1. Scan all members for struct tags and configure if given
2. If type implements "Configure" call it

This dramatically reduces the complexity of Configure methods down to the point where they can be removed completely.


consumer/profiler.go
```diff
@@ -70,35 +70,25 @@ import (
...
+	keepRunning         bool          `config:"KeepRunning"`
 	templates           [][]byte
-	chars               string
-	message             string
-	delay               time.Duration
-	keepRunning         bool
 }
 
-var profilerDefaultCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890 "
+//var profilerDefaultCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890 "
```
msiebeneicher: remove this line if it is unnecessary

core/directproducer.go
```diff
@@ -63,14 +63,13 @@ type DirectProducer struct {
 }
 
 // Configure initializes the standard producer config values.
-func (prod *DirectProducer) Configure(conf PluginConfigReader) error {
-	return prod.SimpleProducer.Configure(conf)
+func (prod *DirectProducer) Configure(conf PluginConfigReader) {
```
msiebeneicher: can you remove this method now?

PR #145, opened 2017-06-19
Config examples: Formatters -> Modulators

ppar:
Updates to configuration examples
- Replace "Formatters:" with "Modulators:" 
- Comment out "Formatter: " since this seems to be obsolete
- Add an example of logging the _GOLLUM_ stream


ppar:
Should be ready for merging

Issue #146, opened 2017-06-21
Gollum shuts down uncleanly


ppar:
At current master (463ce5720a459f885cd5885cbfb2674377d60371), Gollum outputs the following stack trace when shutting down:

```
Panic shutdown: runtime error: invalid memory address or nil pointer dereference
	/Users/pparkkali/src/go/src/github.com/trivago/gollum/vendor/github.com/trivago/tgo/runtime.go:61
	/usr/local/opt/home-brew/Cellar/go/1.8/libexec/src/runtime/asm_amd64.s:514
	/usr/local/opt/home-brew/Cellar/go/1.8/libexec/src/runtime/panic.go:489
	/usr/local/opt/home-brew/Cellar/go/1.8/libexec/src/runtime/panic.go:63
	/usr/local/opt/home-brew/Cellar/go/1.8/libexec/src/runtime/signal_unix.go:290
	/usr/local/opt/home-brew/Cellar/go/1.8/libexec/src/time/sleep.go:75
	/Users/pparkkali/src/go/src/github.com/trivago/gollum/core/logconsumer.go:124
	/Users/pparkkali/src/go/src/github.com/trivago/gollum/coordinator.go:139
	/Users/pparkkali/src/go/src/github.com/trivago/gollum/vendor/github.com/trivago/tgo/runtime.go:97
	/usr/local/opt/home-brew/Cellar/go/1.8/libexec/src/runtime/asm_amd64.s:2197
```

The same stack trace is generated when killing Gollum using Ctrl-C and at the end of a  `consumer.Profiler` run with `KeepRunning` set to `false`.

Example config:

```yaml
# -*- yaml -*-

JunkGenerator:
  Type: "consumer.Profiler"
  Message: "%20s"
  Streams: "junkstream_00"
  KeepRunning: false
  Runs: 5
  Batches: 3
  DelayMs: 100

JunkPrinter:
  Type: "producer.Console"
  Streams: "junkstream_00"
  Modulators:
    - "format.Envelope":
        Prefix: "[JUNK] "

# Print Gollum's application log to stdout/stderr
GollumLogPrinter:
    Type: "producer.Console"
    Console: "stdout"
    #Console: "stderr"
    Streams: "_GOLLUM_"
    Enable: true
    Modulators:
        # Prepend a timestamp
        - "format.Timestamp"
        # Prepend "[_GOLLUM_] " and append a newline
        - "format.Envelope":
            Prefix: "[_GOLLUM_] "
```

[...]

Issue #87, opened 2017-04-03
Use another package for producer.ElasticSearch


arnecls:
The package currently used for producer.ElasticSearch is using a deprecated protocol and hasn't been updated for at least 2 years. We should maybe switch to https://github.com/olivere/elastic which is not as easy to use but supports the full feature set.
This will probably end up to be a complete rewrite of that producer.

msiebeneicher:
PR is ready to merge

PR #150, opened 2017-06-23
Enhancement/switch elasticsearch  package

msiebeneicher:
- Updated dependencies
  - Removed github.com/mattbaird/elastigo package
  - Added gopkg.in/olivere/elastic.v5 package
- New Elasticsearch producer implementation ( #87 )
  - Updated config format
  - Added `SetGzip` feature
  - Removed now unsupported `Connections` feature
  - Changed to batch producer implementation with bulk requests for messages
- Fixed `shutdown panic` from issue #146


msiebeneicher:
This PR now also fix the `shutdown panic` from issue #146

README.md
```diff
@@ -296,6 +296,21 @@ cp contrib_loader.go.dist contrib_loader.go
...
+go build -ldflags='-s -linkmode=internal' -gcflags='-N -l'
+```
+
+With this debug build you are able to start a [Delve](https://github.com/derekparker/delve) remote debugger:
+
+```bash
+# for the gollum arguments pls use this format: ./gollum -- -c my/config.conf
+dlv --listen=:2345 --headless=true --api-version=2 --log exec ./gollum -- -c testing/configs/test_router.conf -ll 3
+```
+
```
ppar: Could add the build and dlv commands to makefile

producer/elasticsearch.go
```diff
@@ -15,297 +15,433 @@
...
 //
-// DayBasedIndex can be set to true to append the date of the message to the
-// index as in "<index>_YYYY-MM-DD". By default this is set to false.
+// Retry/TimeToWaitSec denotes the time in seconds after which a failed dataset will be
+// transmitted again. By default this is set to 3.
 //
-// Servers defines a list of servers to connect to. The first server in the list
-// is used as the server passed to the "Domain" setting. The Domain setting can
-// be overwritten, too.
+// SetGzip enables or disables gzip compression (disabled by default).
```
ppar: This wording is kind of confusing, at least when read without any prior context. What gets compressed and when?

Maybe we could mention that it corresponds directly to the http://godoc.org/gopkg.in/olivere/elastic.v5#SetGzip SetGzip option in the elastic client library (which enables in-transit compression in the protocol, presumably?)

producer/elasticsearch.go
```diff
@@ -15,297 +15,433 @@
...
-	sendIndex := index
-	if prod.dayBasedIndex {
-		sendIndex += msg.GetCreationTime().Format("_2006-01-02")
+func (conn *elasticConnection) isConnected() bool {
+	return conn.isConnectedStatus
+}
+
+func (conn *elasticConnection) connect() error {
+	conf := []elastic.ClientOptionFunc{elastic.SetURL(conn.servers...), elastic.SetSniff(false), elastic.SetGzip(conn.setGzip)}
+	if len(conn.user) > 0 {
```
ppar: Maybe this could be `if len(conn.user) > 0 || len(conn.password)` to support the (rare) case where the username is "" but a password is still set.

ppar:
Looks good to me, added a couple of trivial comments above

README.md
msiebeneicher: not really i guess - problem is the path to your config file which is different for each execution: `-c testing/configs/test_router.conf`

producer/elasticsearch.go
msiebeneicher: i will check that

producer/elasticsearch.go
msiebeneicher: elastic search use a basic auth authentication - i don't see a possibility in elastic search to add credentials without a username: see https://www.elastic.co/guide/en/shield/current/native-realm.html#managing-native-users

msiebeneicher:
@arnecls: Please review the changes later on

Issue #137, opened 2017-06-07
Reduce file polling in consumer.File


arnecls:
In regard to PR #134 and #135 we need to reduce the file polling done by consumer.File.
Either we should more intelligently do the polling or try to utilise fsevents, see https://github.com/fsnotify/fsnotify.

arnecls:
The merge does not implement what is stated above.
As stated in the PRs for 0.4.0 this a symptomatic fix, not addressing the real issue.

msiebeneicher:
How discussed: We implement a "file system event watcher" as separated mode and keep the current `poll mode` as default.

I would not recommend to use the file system event mode for log rotations or heavy frequented file.

PR #160, opened 2017-07-03
Feature/add fs notification mode to file consumer

msiebeneicher:
This PR is related to #137 

- Added watch mode with file system event notifications
- bundled seeker properties


consumer/file.go
```diff
@@ -186,85 +180,98 @@ func (cons *File) initFile() {
...
+			onAfterRead()
 
-			case err == io.EOF:
-				if cons.file.Name() != cons.realFileName {
+		case err == io.EOF:
+			if cons.source.file.Name() != cons.source.realFileName {
+				cons.Logger.Info("Rotation detected")
+				cons.onRoll()
+			} else {
+				newStat, newStatErr := os.Stat(cons.source.realFileName)
```
arnecls: This will still call os.Stat every loop. To avoid this, introduce a timer or counter and only call os.Stat every x.
e.g.
```go
lastStat := time.Now()
// ...
if time.Since(lastStat) > time.Second {
  // ... do os.Stat stuff
  lastStat = time.Now()
}
```

consumer/file.go
```diff
@@ -280,8 +287,186 @@ func (cons *File) Consume(workers *sync.WaitGroup) {
...
+	offset   int64
+}
+
+func newSeeker(conf core.PluginConfigReader) seeker {
+	switch strings.ToLower(conf.GetString("DefaultOffset", fileOffsetEnd)) {
+	default:
+		fallthrough
+	case fileOffsetEnd:
+		return seeker{
+			seek:     2,
```
arnecls: magic numbers, what do they mean?

consumer/file.go
```diff
@@ -280,8 +287,186 @@ func (cons *File) Consume(workers *sync.WaitGroup) {
...
+	w.retry = make(chan int)
+
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		w.logger.Fatal(err)
+	}
+	defer w.close(watcher)
+
+	// watcher init loop
+	go func() {
```
arnecls: that's quite a huge lambda. Should be moved to a separate function

consumer/file.go
```diff
@@ -280,8 +287,186 @@ func (cons *File) Consume(workers *sync.WaitGroup) {
...
+	for w.source.state != fileStateDone {
+		time.Sleep(time.Second)
+	}
+
+	w.logger.Debug("shutdown file watcher ..")
+	close(w.done)
+}
+
+func (w *watcher) watchLoop(watcher *fsnotify.Watcher) {
+	for {
```
arnecls: if this is a select, stopping at done, why the select in the calling function?
This can be moved up.

consumer/file.go
```diff
@@ -280,8 +287,186 @@ func (cons *File) Consume(workers *sync.WaitGroup) {
...
+			if event.Op&fsnotify.Write == fsnotify.Write {
+				w.logger.Debug("modified file: ", event.Name)
+				w.read()
+				break
+			}
+
+			if event.Op&fsnotify.Rename == fsnotify.Rename || event.Op&fsnotify.Remove == fsnotify.Remove {
+				w.logger.WithField("event", event).Debug("file renamed/removed: ", event.Name)
+
+				w.retry <- 1
```
arnecls: this is also not needed when moved to the caller

msiebeneicher:
@arnecls : requested changes updated

consumer/file.go
```diff
@@ -280,8 +296,172 @@ func (cons *File) Consume(workers *sync.WaitGroup) {
...
+}
+
+func (w *watcher) watchLoop(watcher *fsnotify.Watcher) {
+outerLoop:
+	for {
+		if _, err := os.Stat(w.source.realFileName); os.IsNotExist(err) {
+			w.logger.WithField("file", w.source.realFileName).
+				Warning("watched file not exists. retry in 3sec ..")
+			time.Sleep(3 * time.Second)
+			continue outerLoop // retry
```
arnecls: label not needed here

consumer/file.go
```diff
@@ -280,8 +296,172 @@ func (cons *File) Consume(workers *sync.WaitGroup) {
...
+		}
+
+		// read current file state before watching
+		w.read()
+
+		err := watcher.Add(w.source.realFileName)
+		if err != nil {
+			w.logger.Error("error during adding watcher: ", err)
+			time.Sleep(3 * time.Second)
+			continue outerLoop // retry
```
arnecls: label not needed here

consumer/file.go
```diff
@@ -280,8 +296,172 @@ func (cons *File) Consume(workers *sync.WaitGroup) {
...
+		}
+
+	innerLoop:
+		for {
+			select {
+			case event := <-watcher.Events:
+				if event.Op&fsnotify.Write == fsnotify.Write {
+					w.logger.Debug("modified file: ", event.Name)
+					w.read()
+					break
```
arnecls: breaks inner loop - intended?

consumer/file.go
```diff
@@ -280,8 +296,172 @@ func (cons *File) Consume(workers *sync.WaitGroup) {
...
+				}
+
+				if event.Op&fsnotify.Rename == fsnotify.Rename || event.Op&fsnotify.Remove == fsnotify.Remove {
+					w.logger.WithField("event", event).Debug("file renamed/removed: ", event.Name)
+					break innerLoop
```
arnecls: label not needed here

consumer/file.go
```diff
@@ -186,87 +180,109 @@ func (cons *File) initFile() {
...
-					if newStatErr == nil && oldStatErr == nil && !os.SameFile(newStat, oldStat) {
-						cons.Logger.Info("Rotation detected")
-						cons.onRoll()
-					}
-				}
-				spin.Yield()
+func (source *sourceFile) isRotated() bool {
+	if source.file.Name() != source.realFileName {
+		return true
+	} else if time.Since(source.lastStatCheck) > time.Second {
```
arnecls: "else" not needed because of return
This should be a linter warning, too ...?

consumer/file.go
arnecls: make this continue

consumer/file.go
arnecls: Better name? Maybe event loop ..?
IMO it's better to say what "action" you want to break

PR #161, opened 2017-07-03
Parallel enqueue

arnecls:
Allow consumers to enqueue messages by utilising multiple go routines.
This should give a performance boost for consumers with a lot / expensive modulators attached.

Part of this PR are also:
- Remove of the "Instances" plugin parameter (see below)
- Fixes a race condition with stream based metrics

The "Instances" parameter was removed because it was considered dangerous (e.g. when using multiple kafka consumers without a consumer group) and the only reason to use it becomes obsolete with this change.


core/metrics.go
```diff
@@ -126,9 +127,12 @@ func CountFallbackRouters() {
 }
 
 var streamMetrics = map[MessageStreamID]StreamMetric{}
+var streamMetricsGuard = new(sync.Mutex)
 
-// GetSteamMetric return a StreamMetric instance for the given MessageStreamID
-func GetSteamMetric(streamID MessageStreamID) StreamMetric {
+// GetStreamMetric return a StreamMetric instance for the given MessageStreamID
```
ppar: *returns

ppar:
Looks OK to me

consumer/profiler.go
```diff
@@ -196,7 +196,7 @@ func (cons *Profiler) profile() {
 		for i := 0; i < cons.profileRuns && cons.IsActive(); i++ {
 			template := cons.templates[rand.Intn(len(cons.templates))]
 			messageCount++
-			cons.Logger.Debugf("Enqueuing template: '%s'", template)
+			//cons.Logger.Debugf("Enqueuing template: '%s'", template)
```
msiebeneicher: Can you remove this line completely?

core/simpleconsumer.go
```diff
@@ -31,6 +32,8 @@ import (
 //    Enable: true
 //    ID: ""
 //    ShutdownTimeoutMs: 1000
+//    ModulatorQueues: 0
```
msiebeneicher: Rename this setting to `ModulatorRoutines` - That should be easier to understand.
How discussed: Use one queue instead of n to avoid starvation.

PR #163, opened 2017-07-05
fix native kafka producer

msiebeneicher:
* fixed breaking changes in native kafka producer
* provided new dockerfile-dev variant with installed pcap and kafka libs

_Note: only tested by compiling_


arnecls:
This thing requires more refactoring (e.g. key formatter)

PR #164, opened 2017-07-06
JSON to InfluxDB support

mre:
This PR adds support for convert arbitrary JSON payloads into the InfluxDB Line Protocol format.
It has similar functionality like the [logstash influxdb plugin](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-influxdb.html). In combination with the grok formatter it makes migration from logstash to gollum much easier.

#### Implementation details

I tried using the new gollum config format for the plugin configuration.
Namely, I added two struct tags, `config:"Timefield" default:"time"` and `config:"Measurement" default:"measurement"`.
For the two config options "tags" and "ignore" I used a custom hashmap which gets initialized inside the Configure method. This way we can do a fast hashmap lookup which significantly speeds up message parsing.


format/jsontoinflux10.go
```diff
@@ -0,0 +1,175 @@
...
+	err := json.Unmarshal(content, &values)
+
+	if err != nil {
+		format.Logger.Warningf("JSON parser error: %s, Message: %s", err, content)
+		return err
+	}
+
+	var timestamp int64
+	if val, ok := values[format.timeField]; ok {
+		timestamp, err = strconv.ParseInt(val.(string), 10, 64)
```
arnecls: Maybe use `time.Parse` here?
That would allow custom date formats.

format/jsontoinflux10.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Create a set from the ignore array for O(1) lookup
+	ignoreArray := conf.GetStringArray("Ignore", []string{})
+	format.ignore = map[string]bool{}
+	for _, v := range ignoreArray {
+		format.ignore[v] = true
+	}
+}
+
+func (format *JSONToInflux10) escapeMeasurment(value string) string {
```
arnecls: Why not `value []byte`? If these are meant to be "maybe I want to exchange the implementation in the future" wrappers it would make sense to use the datatype as given by the pipeline?
Otherwise these functions don't make sense to me: Don't solve future problems  + the golang inlining is not that great, yet, i.e. this will create unnecessary overhead.

format/jsontoinflux10.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	fields := make(map[string]interface{})
+	tags := make(map[string]interface{})
+	for k, v := range values {
+		_, ignore := format.ignore[k]
+		if ignore {
+			continue
+		}
+		_, isTag := format.tags[k]
```
arnecls: you really don't like the `if x := y; x {` short form, do, you? ;)

format/jsontoinflux10.go
```diff
@@ -0,0 +1,175 @@
...
+	} else {
+		timestamp = time.Now().Unix()
+	}
+
+	var measurement string
+	if val, ok := values[format.measurement]; ok {
+		measurement = val.(string)
+		delete(values, format.measurement)
+	} else {
+		return fmt.Errorf("Required field for measurement (%s) not found in payload", format.measurement)
```
arnecls: I would turn this around to improve readability / put more emphasis on the early out.
```go
val, ok := values[format.measurement]
if !ok {
    return fmt.Errorf("Required field for measurement (%s) not found in payload", format.measurement)
)
measurement := val.(string)
delete(values, format.measurement)
```

format/jsontoinflux10.go
mre: Clarified offline.
`json.Unmarshal()` returns a `map[string]interface{}`, so I need to add a type assertion when calling `escape*`, for example `format.escapeMetadata(v.(string))`. (Note that `format.escapeMetadata(string(v))` does not work. It would result in an error: ["cannot convert v (type interface{}) to type string: need type assertion"](https://stackoverflow.com/a/14289568/270334))

format/jsontoinflux10.go
mre: Done. Also added support for custom time formats.

format/jsontoinflux10.go
mre: Great idea! Fixed.

format/jsontoinflux10.go
mre: Yeah, I tend to forget that one. It's more idiomatic. Fixed.

arnecls:
LGTM

Issue #100, opened 2017-04-24
Update architecture documentation


arnecls:
The current architecture documentation (read the docs RST files) are still from the 0.3.x era.
Change them according to the new structure and add FAQ about architecture by plugin developers.

- [x] High level architecture of message passing
- [x] Explaining the "stream width" (ModulatorQueues)
- [x] Purpose of the different plugin types
- [x] Components, purposes and type hierarchy
- [x] Basic implementation examples

Update - Open tasks before v0.5.0 release:

- [x] Finalise release notes
- [x] Describe Metadata handling in Terminology
- [x] Add Guide to update plugins from v0.4 to v0.5
- [x] Run gollum benchmark of v0.4 VS v0.5
- [x] Add docs for metrics
- [x] Add docs for healthchecks 
- [ ] Describe buffered- and batched-producer components (optional)

ppar:
For the end user docs, rIght now there's a lot of duplication between comments in the source code and the .rst files (namely plugins' config settings), and also between the .rst files themselves (a lot of them begin with the same boilerplate text). 

How can we avoid this? Can we autogenerate some/all of the docs from the sources and use some include mechanism for the repeated parts?

A visual architecture overview would be really useful for the end user as well, to explain how the plugins do (and don't) interact, and where to implement specific functionality (e.g. that messages can only be modified at the consumer stage and not during routing and the retry logic (#99) rationale behind this).

arnecls:
The rst files are generated by an internal tool from the source code comments.

ppar:
Which one and how? 

I couldn't find any, the top level makefile doesn't have a target for this and docs/Makefile just seems to generate docs from the .rsts

arnecls:
It's called gollum-docs and it's one big, ugly file with a lot of unchecked assumptions in it.
It's on the list of the "things-to-clean-up-and-release".
Probably a good thing to do along with this task.

PR #166, opened 2017-07-09
Cloudwatch Logs producer

luqasz:
Hi

I'd like to add cloudwatch logs producer to gollum. AWS puts some limitations on this service which are written in this plugin constants. I have couple of problems here.

- Which type of base producer to use. AWS limits uploads to 5 / second.
- I'd like to use templates for outgoing messages. Some one might want to send only syslog message itself and ignore stuff like priority and facility. I need to parse messages into a struct to be able to use templates. Any formatter can do that ?

Can you please give some advice / guide ?


producer/cloudwatch_logs.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func init() {
+	core.TypeRegistry.Register(CloudwatchLogs{})
+}
+
+// Configure initializes this producer with values from a plugin config.
+func (prod *CloudwatchLogs) Configure(conf core.PluginConfigReader) error {
+	prod.SetStopCallback(prod.flush)
+	stream := conf.GetString("stream", "")
```
arnecls: Consider using struct tags for stream and group (non-pointer) and only leave the checks on empty in the config method

producer/cloudwatch_logs.go
```diff
@@ -0,0 +1,164 @@
...
+		LogEvents:     logevents,
+		LogGroupName:  prod.group,
+		LogStreamName: prod.stream,
+		SequenceToken: prod.token,
+	}
+	// When rejectedLogEventsInfo is not empty, app can not
+	// do anything reasonable with rejected logs. Ignore it.
+	// Meybe expose some statistics for rejected counters.
+	resp, err := dst.svc.PutLogEvents(prod.cloudwatchLogsParams)
+	if err == nil {
```
arnecls: if err != nil there should be an error message logged

producer/cloudwatch_logs.go
```diff
@@ -0,0 +1,164 @@
...
+	eventSizeOverhead = 26
+	// DescribeLogStreams transactions/second.
+	describeLogstreamsDelay = 200 * time.Millisecond
+	// PutLogEvents 5 requests/second/log stream.
+	putLogEventsDelay = 200 * time.Millisecond
+)
+
+type CloudwatchLogs struct {
+	core.BufferedProducer `gollumdoc:"embed_type"`
+	stream                *string
```
arnecls: Make these strings non-pointer.
- It will make it easier to process your config
- Strings are not scattered around in memory (cache locality, memory fragmentation, etc - won't hurt you here but in general I would consider it a good practice)

producer/cloudwatch_logs.go
```diff
@@ -0,0 +1,164 @@
...
+
+// Create log group and stream. If an error is returned, PutLogEvents cannot succeed.
+func (prod *CloudwatchLogs) create() (err error) {
+	err = prod.createGroup()
+	if err != nil {
+		return
+	}
+	err = prod.createStream()
+	return
+}
```
arnecls: Fine like this, but you could also do it like this:
```go
func (prod *CloudwatchLogs) create() error {
    if err := prod.createGroup(); err != nil {
        return err
    }
    return prod.createStream()
}
```
I'll leave it up to you which version you like more.

producer/cloudwatch_logs.go
```diff
@@ -0,0 +1,164 @@
...
+			return nil
+		}
+	}
+	return err
+}
+
+// http://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_CreateLogStream.html
+func (prod *CloudwatchLogs) createStream() error {
+	params := &cloudwatchlogs.CreateLogStreamInput{
+		LogGroupName:  &prod.group,
```
arnecls: This should not compile ...?
`LogGroupName` is a `*string`, not a `**string`.

producer/cloudwatch_logs.go
```diff
@@ -0,0 +1,164 @@
...
+	// When rejectedLogEventsInfo is not empty, app can not
+	// do anything reasonable with rejected logs. Ignore it.
+	// Meybe expose some statistics for rejected counters.
+	resp, err := dst.svc.PutLogEvents(prod.cloudwatchLogsParams)
+	if err == nil {
+		prod.cloudwatchLogsParams.SequenceToken = resp.NextSequenceToken
+	}
+}
+
+func (prod *Console) Produce(workers *sync.WaitGroup) {
```
arnecls: this should be `func (prod *CloudwatchLogs) ...`

producer/cloudwatch_logs.go
```diff
@@ -0,0 +1,164 @@
...
+	defer prod.WorkerDone()
+	prod.AddMainWorker(workers)
+	prod.MessageControlLoop(prod.printMessage)
```
arnecls: printMessage function does not exist.
You probably want to call `upload` here?
If you use the buffered producer you will only get single messages though.
If you want to send batches of messages you should have a look at `core.BatchProducer`

producer/cloudwatch_logs.go
```diff
@@ -0,0 +1,164 @@
...
+import (
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
+	"github.com/trivago/gollum/core"
+	"time"
+)
+
+/*
```
arnecls: Please use `//` when doing comments.
Some people use `/* .. */` to temporarily remove code and its super annoying if you have block comments inside your code that force you to create multiple blocks.

producer/cloudwatch_logs.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	prod.stream = &stream
+	prog.group = &group
+	return nil
+}
+
+// Put log events and update sequence token.
+// Possible errors http://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html
+func (prod *CloudwatchLogs) upload() {
+	logevents := make([]*cloudwatchlogs.InputLogEvent, 0, len(events))
```
arnecls: where does `events` come from?

producer/cloudwatch_logs.go
luqasz: I've created them as pointers to easily set them in upload params.

luqasz:
Where should I put code to:
- get instance metadata (which will communicate with AWS external resources)
- get token (which will communicate with AWS external resources)

It should be launched before first upload to cloudwatch. Is `Produce()` a good place ?

producer/cloudwatchlogs.go
```diff
@@ -0,0 +1,147 @@
...
+	// A batch of log events in a single PutLogEvents request cannot span more than 24 hours.
+	maxBatchTimeSpan = 24 * time.Hour
+	// How many bytes to append to each log event.
+	eventSizeOverhead = 26
+	// DescribeLogStreams transactions/second.
+	describeLogstreamsDelay = 200 * time.Millisecond
+	// PutLogEvents 5 requests/second/log stream.
+	putLogEventsDelay = 200 * time.Millisecond
+)
+
```
msiebeneicher: Please provide a documentation and a config example . Look here for an example: https://github.com/trivago/gollum/blob/master/producer/elasticsearch.go#L31

producer/cloudwatchlogs.go
```diff
@@ -0,0 +1,147 @@
...
+type CloudwatchLogs struct {
+	core.BufferedProducer `gollumdoc:"embed_type"`
+	stream                string `config:"Stream" default:""`
```
arnecls: "Stream" is quite close to the standard "Streams".
As by line 93 I would rename this to "LogStreamPrefix". Avoids ambiguity in the config.

producer/cloudwatchlogs.go
```diff
@@ -0,0 +1,147 @@
...
+	group                 string `config:"Group" default:""`
```
arnecls: To align with my previous proposal I would rename this to "LogGroup".

producer/cloudwatchlogs.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	if conf.GetString("group", "") == "" {
+		prod.Logger.Error("group name can not be empty")
+	}
+}
+
+// Put log events and update sequence token.
+// Possible errors http://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html
+func (prod *CloudwatchLogs) upload(msg *core.Message) {
+	logevents := make([]*cloudwatchlogs.InputLogEvent, 0)
```
arnecls: You create logevents here and upload them later, but there is no data added to it ...?
Also this looks like you actually want to upload more than one event at a time.
As of this I would suggest you base this plugin on core.BatchProducer which gives you a list of messages instead of just one message.

I would suggest to have a look at producer.Kinesis for that.

producer/cloudwatchlogs.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Configure initializes this producer with values from a plugin config.
+func (prod *CloudwatchLogs) Configure(conf core.PluginConfigReader) {
+	if conf.GetString("stream", "") == "" {
+		prod.Logger.Error("stream name can not be empty")
+	}
+	if conf.GetString("group", "") == "" {
+		prod.Logger.Error("group name can not be empty")
+	}
+}
```
arnecls: prod.service is not initialized at all.
If you need help about creation of the session and config object have a look at producer.Kinesis, producer.S3 or consumer.Kinesis (and of course the AWS SDK docs)

arnecls:
> Which type of base producer to use. AWS limits uploads to 5 / second.
As written in the last comment - for AWS a core.BatchProducer is normally the best option as you get a batch of messages instead of single messages. As an advice - have a look at the other AWS producers (Kinesis and S3). Most of the common stuff is in there.

> I'd like to use templates for outgoing messages. Some one might want to send only syslog message itself and ignore stuff like priority and facility. I need to parse messages into a struct to be able to use templates. Any formatter can do that?

Normally we use TextToJSON or write a custom formatter for that. After that you can use the different JSON processors (processJSON, templateJSON) to modify your message. This process will change with 0.6.0 (see #171) but for now this is how it works.

Also - please test your plugin on AWS and notify us when this actually works.
Until then I will mark this as "work in progress".

luqasz:
Thx for your feedback. I will let you know when I am finished.

luqasz:
How can I set default max batch size ? In kinesis producer dosc state that `BatchMaxMessages` is set to 500 but there is no code for that in `producer/kinesis.go`.

msiebeneicher:
The batch settings comes from the embedded batchedproducer.
Here the embed producer: https://github.com/trivago/gollum/blob/master/producer/kinesis.go#L98
And here the settings comes from: https://github.com/trivago/gollum/blob/master/core/batchedproducer.go#L44

But the docs are not correct. The new configuration have to look like 
``` yaml
Batch:
  MaxCount: 8192
  FlushCount: 4096
  TimeoutSec: 5
```

We will update all docs by #100

luqasz:
Ok. Is there any way to set this in code ? I would like to set max limits based on AWS limits so they will not be exceeded by user configuring gollum.

arnecls:
You can do so in the configure method of the plugin.
But as far as I can see this will not be a fun/easy task at the moment.
The configuration order is member-by-member:

- Tags
- If there is a configure method, call it (although currently the member needs to be public for that)

So the last method called is the configure method of your producer.
You could create a new BatchedProducer there - after checking the constraints - and overwrite the one created, but that feels bollocks.
I guess we need to extend the BatchedProducer to allow constraints. @msiebeneicher any suggestions?

arnecls:
We discussed this issue a little bit. We came to the point that it is best to check for the limits in the configure method of your plugin and if a limit is reached push an error.
This will cause gollum to not start and an error to be displayed.

Example:
```go
func (prod *YourProducer) Configure(conf PluginConfigReader) {
    if prod.batchMaxCount > CountLimit {
        conf.Errors.Pushf("Batch/MaxCount must be below %d", CountLimit)
    }
}
```

luqasz:
I am ready. I didn't write any integration test yet. Any hints where should I start and/or how it should look like ?

msiebeneicher:
Integration test are currently pretty hard for aws producers / consumers because of missing credentials for travis.

I'm still confused that the unit test are failing in your PR. Maybe you need to reintegrate the master again?

luqasz:
Only failing test is this:
```=== RUN   TestProducerInterface
time="2017-07-27T21:26:02+02:00" level=error msg="No stream configuration found. Please check your config." PluginID=prod0 PluginType=producer.ElasticSearch
--- FAIL: TestProducerInterface (2.12s)
	producers_test.go:52: Failed to create producer producer.CloudwatchLogs: LogStream can not be empty, LogGroup can not be empty
FAIL```

**Comment by msiebeneicher** [comment]
@luqasz: 

1) The unit test fails because the `TestProducerInterface` try to instantiate all producers.
Because of the empty `LogStream` and `LogGroup` settings we get this error.

To fix this i would recommend this change to fix the current issue:

```golang
// Configure initializes this producer with values from a plugin config.
func (prod *AwsCloudwatchLogs) Configure(conf core.PluginConfigReader) {
	if prod.stream == "" {
		prod.Logger.Error("LogStream can not be empty")
	}
	if prod.group == "" {
		prod.Logger.Error("LogGroup can not be empty")
	}
	if conf.GetInt("Batch/MaxCount", maxBatchEvents) > maxBatchEvents {
		conf.Errors.Pushf("Batch/MaxCount must be below %d", maxBatchEvents)
	}
}
```

2) Please check the new guidelines of how we document the plugins now: http://gollum.readthedocs.io/en/latest/src/instructions/writingPlugins/docs.html 
It is important to use the correct format because we auto-generate the documentation out of the code. 

In this case the doc part should look like this example:
```golang
// AwsCloudwatchLogs producer plugin
//
// The AwsCloudwatchLogs producer plugin sends messages to
// AWS Cloudwatch Logs service.
//
// Parameters
//
// - LogStream: xxxx
//
// - LogGroup: xxx
//
// Examples
//
// This example ...:
//
//	CwLogs:
//		Type: AwsCloudwatchLogs:
//			LogStream: stream_name
//			LogGroup: group_name
//    Credential:
//      Type: shared
//
```

After the requested changes this PR should be mergeable! :)
thx for your work! 👍

luqasz:
I will catch you later. I am busy now. Thx for reply.

luqasz:
Hi. Do you need any thing else from this PR ?

msiebeneicher:
@luqasz : Sorry for the delay! We will merge this PR after the final v0.5.0 release. Last time I saw no further issues with this PR - so i think we are able to merge this directly after the release.

luqasz:
I've allowed myself to add a separate function to handle upload results. Usefull when group / stream was removed during program runtime. Gollum will create group / stream (without uploading batch) so next uploads will succeed.

luqasz:
@arnecls  @msiebeneicher  Do you need anything from me to merge this PR ?

arnecls:
LGTM. If @msiebeneicher does not find anything I'll do the merge on Wednesday when I'm back at the office.

arnecls:
Merged.
Thank you for contributing and for your patience 🍺

PR #168, opened 2017-07-10
Decrease grok verbosity in case of parse errors

mre:
Turns out the grok formatter can become quite verbose when handling lots of invalid data.
By setting the log level to "warning", we at least allow the user to set the log level to 0 and ignore the messages.


arnecls:
LGTM

PR #170, opened 2017-07-11
Fix for #103 Enabled shared config sessions for all AWS related plugins

### LINKED ISSUE #103: S3 Producer doesn't support AWS role switching
When using AWS STS credentials, `producer/s3.go` does not know how to switch to the required AWS STS role. 

Given a Gollum configuration:

```
"StdIn":
    Type: "consumer.Console"
    Streams: "console"
    Fuse: "file"

"S3Out":
    Type: "producer.S3"
    Region: "eu-central-1"
    CredentialType: "shared"
    CredentialFile: "/home/myusername/.aws/credentials"
    CredentialProfile: "test_profile"
    StreamMapping:
        "console": "my-testbucket/console"
        "*" : "my-testbucket/asterisk"
```

And an `~/.aws/credentials` file:

```
[test_profile]
aws_access_key_id = AKI.....
aws_secret_access_key = asdfgh1234
source_profile = default
role_arn = arn:aws:iam::123456789:role/foo-bar
region = eu-central-1
output = table
```

Gollum will simply try to connect using the default profile of that user and ignores the `role_arn` setting. 

There seems to be support for assuming roles in the AWS Go SDK:

https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/stscreds/assume_role_provider.go

So it could be useful to add support for this in Gollum.

arnecls:
This change will allow gollum to be started like this:

```bash
AWS_DEFAULT_PROFILE=foobar gollum -c ...
```

This fixes #103 with a known way to change roles.

### DIFF
--- a/consumer/kinesis.go
+++ b/consumer/kinesis.go
@@ -296,7 +296,14 @@ func (cons *Kinesis) processShard(shardID string) {
 }
 
 func (cons *Kinesis) connect() error {
-	cons.client = kinesis.New(session.New(cons.config))
+	session, err := session.NewSessionWithOptions(session.Options{
+		Config:            *cons.config,
+		SharedConfigState: session.SharedConfigEnable,
+	})
+	if err != nil {
+		return err
+	}
+	cons.client = kinesis.New(session)
 
 	// Get shard ids for stream
 	streamQuery := &kinesis.DescribeStreamInput{
--- a/consumer/kinesis.go
+++ b/consumer/kinesis.go
@@ -27,6 +27,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/kinesis"
 	"github.com/trivago/gollum/core"
@@ -117,6 +118,7 @@ type Kinesis struct {
 	sleepTime           time.Duration `config:"QuerySleepTimeMs" default:"1000" metric:"ms"`
 	retryTime           time.Duration `config:"RetrySleepTimeSec" default:"4" metric:"sec"`
 	shardTime           time.Duration `config:"CheckNewShardsSec" default:"0" metric:"sec"`
+	assumeRole          string        `config:"Credential/AssumeRole"`
 	running             bool
 }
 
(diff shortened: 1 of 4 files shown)


arnecls:
Not tested on AWS, yet.

msiebeneicher:
Change works for s3 producer as expected with example config from @ppar :

```yaml
"StdIn":
    Type: "consumer.Console"
    Streams: "console"
    Fuse: "file"

"S3Out":
    Type: "producer.S3"
    Region: "eu-central-1"
    CredentialType: "shared"
    CredentialFile: "/home/myusername/.aws/credentials"
    CredentialProfile: "test_profile"
    StreamMapping:
        "console": "my-testbucket/console"
```

ppar:
Looks good. We should also mention this in the affected plugins' documentation (at least w/ a short note so that we'll remember to include it in the upcoming documentation rewrite).

mahdi-hdi:
It doesn't work when you don't have an  `aws_access_key_id` or `aws_secret_access_key` in the config files (`~/.aws/credentials` or `~/.aws/config`). 
A use case is when you run an EC2 instance. You usually don't put any api_key in the instance, since it takes permission directly from AWS on startup so it's not necessary to provide any API key or even the `~/.aws/confg` file. In this case, any kinds of creating session or credential based on `~/.aws/confg` doesn't work because having the `aws_access_key_id` is a requirement for it.

arnecls:
Hmm - in that case `CredentialType: "environment"` should work, shouldn't it?

mahdi-hdi:
Based on the environment you are able to just load following variables:
https://github.com/aws/aws-sdk-go/blob/master/aws/session/env_config.go#L97
which is nothing related to assuming a role.

Moreover, by taking a look at how it merges different session configurations (file, env, ...) will find out assuming role without key and secret should be done (I guess) by [`stscreds.AssumeRoleProvider`](http://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds/)

arnecls:
@mahdi-hdi like this (see last commit)?

mahdi-hdi:
Thanks, @arnecls. Now it looks good to me.

arnecls:
Ok, @mahdi-hdi, @msiebeneicher did you test the latest version already on AWS?
If not I will do so tomorrow.

mahdi-hdi:
Nope, will test it today.

arnecls:
There is still some issues with the AssumeRole setting, but this looks more like an SDK related error than something broken on our side. Furthermore @mahdi-hdi got it running with Kinesis, so I assume it _should_ work.
If it does not, we need another fix. /merged

Issue #172, opened 2017-07-11
Syslog consummer panics


luqasz:
Hi.

I've noticed that syslog parsing crashes while message is sent:
```
[2017-07-11 21:41:53 CEST]  INFO A wildcard stream configuration only affects the wildcard stream, not all routers PluginID=_generated_stream_* PluginType=router.Broadcast
[2017-07-11 21:41:53 CEST]  INFO We be nice to them, if they be nice to us. (startup)
panic: runtime error: index out of range

goroutine 35 [running]:
github.com/trivago/gollum/core.(*SimpleConsumer).directEnqueue(0xc4200e8540, 0xc420c82000)
	/Users/lkostka/workspace/golang/src/github.com/trivago/gollum/core/simpleconsumer.go:250 +0x2af
github.com/trivago/gollum/core.(*SimpleConsumer).(github.com/trivago/gollum/core.directEnqueue)-fm(0xc420c82000)
	/Users/lkostka/workspace/golang/src/github.com/trivago/gollum/core/simpleconsumer.go:93 +0x34
github.com/trivago/gollum/core.(*SimpleConsumer).EnqueueWithMetadata(0xc4200e8540, 0xc420c6a090, 0x2a, 0x30, 0x0)
	/Users/lkostka/workspace/golang/src/github.com/trivago/gollum/core/simpleconsumer.go:204 +0x9e
github.com/trivago/gollum/core.(*SimpleConsumer).Enqueue(0xc4200e8540, 0xc420c6a090, 0x2a, 0x30)
	/Users/lkostka/workspace/golang/src/github.com/trivago/gollum/core/simpleconsumer.go:198 +0x52
github.com/trivago/gollum/consumer.(*Syslogd).Handle(0xc4200e8540, 0xc420c6a060, 0x4c, 0x0, 0x0)
	/Users/lkostka/workspace/golang/src/github.com/trivago/gollum/consumer/syslogd.go:114 +0x20b
github.com/trivago/gollum/vendor/gopkg.in/mcuadros/go-syslog%2ev2.(*Server).parser(0xc420c10000, 0xc420c56000, 0x4c, 0x10000, 0xc420c66020, 0xf, 0x0, 0x0)
	/Users/lkostka/workspace/golang/src/github.com/trivago/gollum/vendor/gopkg.in/mcuadros/go-syslog.v2/server.go:270 +0x2be
[...]

Issue #154, opened 2017-06-28
Testconfig mode (-tc / -testconfig) doesn't work


ppar:
Gollum has a mode to "test" a configuration file (for what? syntax errors or also plugin startup?), but all of the invocations below just produce the "usage" message and exit. 

./gollum -tc -c config/http_auth_receive.conf
./gollum -testconfig config/http_auth_receive.conf
./gollum -ll 3 -tc -c config/http_auth_receive.conf

./gollum -tc config/http_auth_receive.conf
./gollum -testconfig -c config/http_auth_receive.conf
./gollum -ll 3 -tc config/http_auth_receive.conf

arnecls:
The "tc" should do two things
a) check if the yaml is parsable (which it does)
b) check if all plugins can be instantiated correctly without really activiating them (which it does not)

So this should be about properly implementing b)

ppar:
c) Print an unambigious result instead of the usage instructions

Issue #151, opened 2017-06-27
Panic when trying to use file consumer without stream configuration


mre:
Gollum version: [Latest master](https://github.com/trivago/gollum/commit/463ce5720a459f885cd5885cbfb2674377d60371), compiled with `make`
Operating system: Mac OS Sierra 10.12.5

**Contents of file: gollum.conf:**
```yaml
input:
    Type: consumer.File
    File: foo
    DefaultOffset: "Oldest"

output:
   Type: producer.Console
   Streams:
    - test
```

**Contents of file: foo:**
```foo```

**Command:**
```./gollum -c gollum.conf```

**Expected behavior:**
Show error message. Something like:
"Missing 'Streams' parameter in plugin `input`. At least one stream must be defined for every plugin."

**Actual behavior:**
Runtime panic.
```
Panic shutdown: runtime error: index out of range
	.../go/src/github.com/trivago/gollum/vendor/github.com/trivago/tgo/runtime.go:61
	/usr/local/opt/go/libexec/src/runtime/asm_amd64.s:514
	/usr/local/opt/go/libexec/src/runtime/panic.go:489
	/usr/local/opt/go/libexec/src/runtime/panic.go:28
	.../go/src/github.com/trivago/gollum/core/simpleconsumer.go:216
	.../go/src/github.com/trivago/gollum/core/simpleconsumer.go:182
	.../go/src/github.com/trivago/gollum/consumer/file.go:121
	.../go/src/github.com/trivago/gollum/consumer/file.go:188
	.../go/src/github.com/trivago/gollum/vendor/github.com/trivago/tgo/tio/bufferedreader.go:267
	.../go/src/github.com/trivago/gollum/consumer/file.go:231
```

**Fix:**
Define at least one stream.
```
input:
    Type: consumer.File
    File: foo
    DefaultOffset: "Oldest"
    Streams:
      - test

output:
   Type: producer.Console
   Streams:
    - test
```
Have not tested, but maybe the same bug occurs in other plugins.

ppar:
OTOH, in some plugins, missing parameters simply cause an Error-level entry in the log, but they don't abort startup. It would be nice to harmonize this behavior. Maybe have `Configure()` in the plugins (`interface Configurable`) return an error value etc.

mre:
True. I'd rather err on the side of caution here and throw and error when the configuration is incomplete in any way. Consistent error behavior would be a big plus.

PR #174, opened 2017-07-12
Fix for #151 config validation on startup

arnecls:
Fix for #151, #172, fixing parts of #154
Configuration error handling and logging was broken in several ways.
The following fixes are included:

- config.Validate is now called again during startup (related to #154)
- "Streams" is now mandatory for consumers and producers
- "Stream" is now mandatory for routers
- fixed missing log messages during shutdown due to a missing purge
- fixed coloring of log messages during startup due to setting the formatter to late
- having no consumer/producers leads to shutdown
- proper use of logrus.WithError wthin main package

### DIFF
--- a/coordinator.go
+++ b/coordinator.go
@@ -130,7 +130,7 @@ func (co *Coordinator) StartPlugins() {
 	if core.StreamRegistry.IsStreamRegistered(core.LogInternalStreamID) {
 		// The _GOLLUM_ stream has listeners, so use LogConsumer to write to it
 		if *flagLogColors != "always" {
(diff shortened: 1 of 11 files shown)


arnecls:
TBD: Streams is actually a requirement made by core.SimpleProducer / core.SimpleConsumer. As of this I would normally check it in the Configure methods of these files but as this is a structural requirement I moved it to the coordinator.
Not quite sure if this is a good idea or not.

arnecls:
I decided to implement point 2 and 3 of my previous comment.
Still undecided about point 1.

Issue #173, opened 2017-07-12
Make producer.S3 work like producer.File


arnecls:
Producer.File cannot utilize producer.Batch as it handles multiple Files in multiple batches.
The first task is about to generalize this behavior for other producers to use.
Producer.S3 concept wise _should_ do the same as producer.File, just on a remote file that happens to be located on S3. So it should behave like producer.File, too. Furthermore there are several conceptual error in the current implementation (like a delimiter that should be a modulator) that need to be fixed.

Tasks:
- Change producer.File to be based on core.DirectProducer
- Refactor producer.fileState so that file is replaced with a file interface instead (open,close,write,stats)
- Move all pruning related functions to a separate struct, no generalization necessary, yet
- Move the current implementation of producer.S3 to producer.deprecated.s3
- Rewrite S3 to behave like producer file (multiple files, no temporary storage, partial updates, etc.)

PR #177, opened 2017-07-14
#173 refactored file producer for s3 producer reusage [PART 1]

msiebeneicher:
Here the first PR for task #173 

This PR mostly contain the refactoring of the file producer.

### DIFF
--- a/core/components/batchedWriterAssembly.go
+++ b/core/components/batchedWriterAssembly.go
@@ -34,7 +34,6 @@ type BatchedWriterAssembly struct {
 }
 
 // BatchedWriter is an interface for different file writer like disk, s3, etc.
-// TODO: Replace methods with os.FileInfo interface
 type BatchedWriter interface {
 	io.WriteCloser
 	Name() string // base name of the file/resource
--- a/core/components/batchedWriterAssembly.go
+++ b/core/components/batchedWriterAssembly.go
@@ -12,20 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package core
+package components
 
 import (
 	"github.com/sirupsen/logrus"
+	"github.com/trivago/gollum/core"
 	"io"
 	"time"
 )
 
 // BatchedWriterAssembly is a helper struct for io.Writer compatible classes that use batch directly for resources
 type BatchedWriterAssembly struct {
-	Batch           MessageBatch // Batch contains the MessageBatch
-	Created         time.Time    // Created contains the creation time from the writer was set
+	Batch           core.MessageBatch // Batch contains the MessageBatch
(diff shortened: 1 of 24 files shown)


msiebeneicher:
There is now an issue in the "AggregatePipeline". I will investigate the problem next week - i guess that is related with some configs

core/components/batchedWriterAssembly.go
```diff
@@ -0,0 +1,112 @@
...
+	writer          BatchedWriter
+	assembly        core.WriterAssembly
+	flushTimeout    time.Duration // max sec to wait before a flush is aborted
+	batchTimeout    time.Duration // max sec to wait before batch will flushed
+	batchFlushCount int
+	logger          logrus.FieldLogger
+}
+
+// BatchedWriter is an interface for different file writer like disk, s3, etc.
+// TODO: Replace methods with os.FileInfo interface
```
arnecls: I think this interface is fine as it is.
os.FileInfo introduces functions not required by the functions using BatchedWriter.
Function that are not used should not be part if the interface.

core/components/batchedWriterAssembly.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// SetWriter set a BatchedWriter interface implementation
+func (bwa *BatchedWriterAssembly) SetWriter(writer BatchedWriter) {
+	bwa.writer = writer
+	bwa.Created = time.Now()
+}
+
+// UnsetWriter unset the current writer
+func (bwa *BatchedWriterAssembly) UnsetWriter() {
```
arnecls: Does this function have to be public? Do we even require it?

producer/file.go
```diff
@@ -165,209 +140,222 @@ func (prod *File) Configure(conf core.PluginConfigReader) {
...
+	for _, batchedFile := range prod.files {
+		batchedFile.FlushOnTimeOut()
 	}
 }
 
 func (prod *File) writeMessage(msg *core.Message) {
-	streamMsg := msg.Clone()
-
-	state, err := prod.getFileState(streamMsg.GetStreamID(), false)
+	batchedFile, err := prod.getBatchedFile(msg.GetOrigStreamID(), false)
```
arnecls: This should be GetStreamID as the "last" stream should define the file name

core/components/batchedWriterAssembly.go
msiebeneicher: This method was in use before i implemented the `GetWriterAndUnset()`. So currently i can change they to private but this is a bit "inconsistent" if you only provide a `GetWriter()` and `GetWriterAndUnset()` method. In my opinion the interface is now cleaner.

producer/file.go
msiebeneicher: right - i will change that

Issue #132, opened 2017-05-26
kinesis-consumer: race condition on shard offset map


mahdi-hdi:
In the Kinesis consumer, we have a race condition on the offsets map (kinesis.offsets) almost in everywhere. For example in:
- [line 299](https://github.com/trivago/gollum/blob/master/consumer/kinesis.go#L299) 
- [line 228](https://github.com/trivago/gollum/blob/master/consumer/kinesis.go#L228)
- and more

From my point of view adding a lock to make it mutual exclusive is not a good idea because it has an impact on performance mainly [line 299](https://github.com/trivago/gollum/blob/master/consumer/kinesis.go#L299). A better solution might be in the way that processShard just work with the pointer of its offset, not whole offsets map structure. 

Moreover, the current implementation of [processShard](https://github.com/trivago/gollum/blob/master/consumer/kinesis.go#L253) and [updateShards](https://github.com/trivago/gollum/blob/master/consumer/kinesis.go#L349)  doesn't handle resharding properly. For example, if a shard expired and get deleted (either splitShard or mergeShards) the goroutine(processShard) which was responsible for it won't be ended and try to read from a shard that doesn't exist anymore.

arnecls:
Mutexes in Go are very lightweight, so unless you don't measure a performance problem they are good to go. However - I agree with your idea to couple the offset with the function because avoiding a mutex always wins.

arnecls:
Passing the offset to the process function does not work well because the offsets still need to be gathered to be written to the offset file.

If the offsets are stored as local variables there is no way of querying them from outside. You would have to push the results to another function for storage which will again create a congestion point, i.e. you will need a mutex there.
So the best approach IMHO is to add an RWMutex here.

mahdi-hdi:
:+1:

PR #178, opened 2017-07-17
Fix for #132: Fix a race condition in consumer.Kinesis

arnecls:
As described in #132 this fixes the map access race condition when storing/reading offsets.
@mahdi-hdi Please check if this is working for you.


mahdi-hdi:
It looks good to me, tnx.

PR #182, opened 2017-07-20
#173 make s3 like file producer

msiebeneicher:
This PR is related to #173 :

* Provided new awsS3 producer with multipart upload support
* Moved old S3 producer to contrib/deprecated/
* Fixed bug in rotation component for "rotateAt" settings
* Updated dependencies

Open todo: 
- [x] Check need of sync.mutex in s3Writer


core/components/awsMultiClient.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// GetS3Client returns a s3.S3 client
+func (client *AwsMultiClient) GetS3Client() (*s3.S3, error) {
+	sess, err := client.newSessionWithOptions()
+	if err != nil {
+		return nil, err
+	}
+
+	return s3.New(sess, client.config), nil
```
arnecls: You could do it like this:
```go
func (client * AwsMultiClient) NewClient(builder func(*session.Session, *aws.Config)) (interface{}, error) {
	sess, err := client.newSessionWithOptions()
	if err != nil {
		return nil, err
	}
	
	return builder(sess, client.config), nil
}
```

core/components/awsMultiClient.go
```diff
@@ -0,0 +1,136 @@
...
+		return credentials.NewStaticCredentials(cred.staticID, cred.staticSecret, cred.staticToken)
+
+	case credentialTypeShared:
+		return credentials.NewSharedCredentials(cred.sharedFile, cred.sharedProfile)
+
+	case credentialTypeNone:
+		return credentials.AnonymousCredentials
+
+	default:
+		panic(fmt.Sprintf("Unknown CredentialType: %s", cred.credentialType))
```
arnecls: return an error instead, please, so a plugin's configure method can in the end work with this.

core/components/batchedWriterAssembly.go
```diff
@@ -109,3 +110,55 @@ func (bwa *BatchedWriterAssembly) FlushOnTimeOut() {
...
+	if time.Since(bwa.Created) >= rotate.Timeout {
+		bwa.logger.Debug("Rotate true: ", "lifetime > timeout setting")
+		return true, nil // ### return, too old ###
+	}
+
+	// RotateAt crossed?
+	if rotate.AtHour > -1 && rotate.AtMinute > -1 {
+		now := time.Now()
+		rotateAt := time.Date(now.Year(), now.Month(), now.Day(), rotate.AtHour, rotate.AtMinute, 0, 0, now.Location())
+
```
arnecls:
```go
if rotateAt.Before(bwa.Created) {
    rotateAt.Add(time.Day)
}
```

core/components/batchedWriterAssembly.go
```diff
@@ -109,3 +110,55 @@ func (bwa *BatchedWriterAssembly) FlushOnTimeOut() {
...
+		if rotateAt.Unix() <= bwa.Created.Unix() && bwa.Created.Sub(rotateAt).Minutes() <= 0 {
```
arnecls:
```go
if now.After(rotateAt) {
    // ...
}
```

core/components/batchedWriterAssembly.go
arnecls: Example:
- rotateAt is set to 11:00
- the file was created at 11:01
- we hit this code at  11:06

As rotateAt lies before created we need to move rotateAt 1 day into the future (don't rotate).
Now if we pass midnight, rotateAt will be after created (different day) and this check will not pass, i.e. no day is added. When we now pass the mark, a rotation will be triggered.

core/components/batchedWriterAssembly.go
arnecls: This is what we want. If we passed rotateAt, trigger a rotate.
This will cause created to be set to a time AFTER rotateAt, so the fix above (add 1 day) will be triggered again.

core/components/batchedWriterAssembly.go
arnecls: Even better:
```go
if rotateAt.Before(bwa.Created) {
    return false, nil
}
```
As rotateAt is in the past (before created = last rotation), we're done here

producer/awsS3.go
```diff
@@ -0,0 +1,281 @@
...
+func (prod *AwsS3) getFinalFileName(baseFileName string) string {
+	fileExt := filepath.Ext(baseFileName)
+	fileName := baseFileName[:len(baseFileName)-len(fileExt)]
+
+	timestamp := time.Now().Format(prod.Rotate.Timestamp)
+	signature := fmt.Sprintf("%s_%s", fileName, timestamp)
+	maxSuffix := uint64(0)
+
+	//todo: get list from aws?
+	/*files, _ := ioutil.ReadDir(streamFile.dir)
```
arnecls: As discussed.
To avoid too many HTTP calls (e.g. multiple rotates per second) we should list the files once at start and keep a "cache" of the latest file name. As writing to the same file/folder from different instances will break, we can be pretty sure we are the only one doing this.

However - this was decided to be too much for 0.5.0, so create an issue for 0.5.x so we support this feature later (because it is a valid thing)

producer/awsS3/batchedFileWriter.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (w *BatchedFileWriter) uploadPartInput() (err error) {
+	if size, _ := w.activeBuffer.Size(); size < 1 {
+		w.logger.Warning("uploadPartInput(): empty buffer - no upload necessary")
+		return
+	}
+
+	//todo: check guard necessity
+	//w.writeGuard.Lock()
```
arnecls: Not required as long as there is no call to Write during Close.

producer/awsS3.go
```diff
@@ -0,0 +1,281 @@
...
+	// Rotate is public to make Pruner.Configure() callable (bug in treflect package)
+	Rotate         components.RotateConfig   `gollumdoc:"embed_type"`
+	AwsMultiClient components.AwsMultiClient `gollumdoc:"embed_type"`
+
+	// configuration
+	batchTimeout      time.Duration `config:"Batch/TimeoutSec" default:"5" metric:"sec"`
+	batchMaxCount     int           `config:"Batch/MaxCount" default:"1000"`
+	batchFlushCount   int           `config:"Batch/FlushCount" default:"500"`
+	batchFlushTimeout time.Duration `config:"Batch/FlushTimeoutSec" default:"0" metric:"sec"`
+	bucket            string        `config:"Bucket" default:""`
```
arnecls: Take this block and move it into a separate type. Make this type a factory for batchedFiles.
```go
type BatchedFileBuilder struct {
	batchTimeout      time.Duration `config:"Batch/TimeoutSec" default:"5" metric:"sec"`
 	batchMaxCount     int           `config:"Batch/MaxCount" default:"1000"`
 	batchFlushCount   int           `config:"Batch/FlushCount" default:"500"`
 	batchFlushTimeout time.Duration `config:"Batch/FlushTimeoutSec" default:"0" metric:"sec"`
}
```

... continued below

producer/awsS3.go
```diff
@@ -0,0 +1,281 @@
...
+	formatString := "%s_%d%s"
+	if prod.Rotate.ZeroPad > 0 {
+		formatString = fmt.Sprintf("%%s_%%0%dd%%s", prod.Rotate.ZeroPad)
+	}
+	return fmt.Sprintf(formatString, signature, int(maxSuffix), fileExt)
+
+}
+
+func (prod *AwsS3) newBatchedWriterAssembly() *components.BatchedWriterAssembly {
+	batchedFile := components.NewBatchedWriterAssembly(
```
arnecls: Move this function to the BatchedFileBuilder type as suggested above.
So you can e.g. call `builder.New(prod)` to get a new batchedFile.

producer/awsS3/batchedFileWriter.go
```diff
@@ -0,0 +1,237 @@
...
+	s3UploadID       *string             // upload id from s3 for active file
+	totalSize        int                 // total size off all writes to this writer (need for rotations)
+	completedParts   []*s3.CompletedPart // collection of uploaded parts
+
+	// need separate byte buffer for min 5mb part uploads.
+	// @see http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html
+	activeBuffer *s3ByteBuffer
+
+	//todo: check necessity
+	writeGuard *sync.Mutex
```
arnecls: remove, see below

producer/awsS3/byteBuffer.go
```diff
@@ -0,0 +1,89 @@
...
+func (buf *s3ByteBuffer) Size() (int, error) {
+	return len(buf.bytes), nil
+}
+
+func (buf *s3ByteBuffer) Sha1() (string, error) {
+	hash := sha1.Sum(buf.bytes)
+	return hex.EncodeToString(hash[:]), nil
+}
+
+/*func (buf *s3ByteBuffer) Compress() error {
```
arnecls: remove, the compression functionality is part of S3 itself

producer/file.go
```diff
@@ -102,7 +101,7 @@ type File struct {
 	batchTimeout      time.Duration `config:"Batch/TimeoutSec" default:"5" metric:"sec"`
 	batchMaxCount     int           `config:"Batch/MaxCount" default:"8192"`
 	batchFlushCount   int           `config:"Batch/FlushCount" default:"4096"`
-	flushTimeout      time.Duration `config:"FlushTimeoutSec" default:"0" metric:"sec"`
+	batchFlushTimeout time.Duration `config:"Batch/FlushTimeoutSec" default:"0" metric:"sec"`
```
arnecls: Like the S3 producer we can use the suggested component factory here

Issue #195, opened 2017-08-02
Race condition in core.PluginRegistry


arnecls:
Not actually reported as a race condition but it is one.
Streams can be generated during runtime, which also means that a router is instantiated at that point. This can lead to the following panic:

```
panic: Plugin id '_GENERATED_orchestra_application_state' must be unique

goroutine 168 [running]:
github.com/trivago/gollum/core.(*streamRegistry).createFallback(0x1eece80, 0xd4774cdf3f9854e3, 0xd4774cdf3f9854e3, 0x1f0a6e0)
	/Users/aclaus/go/src/github.com/trivago/gollum/core/streamregistry.go:193 +0x443
github.com/trivago/gollum/core.(*streamRegistry).GetRouterOrFallback(0x1eece80, 0xd4774cdf3f9854e3, 0xd4774cdf3f9854e3, 0x716)
	/Users/aclaus/go/src/github.com/trivago/gollum/core/streamregistry.go:217 +0xdd
github.com/trivago/gollum/router.(*Metadata).Enqueue(0xc420c14460, 0xc421f39680, 0x20, 0xc421e4a780)
	/Users/aclaus/go/src/github.com/trivago/gollum/router/metadata.go:61 +0xcb
github.com/trivago/gollum/core.Route(0xc421f39680, 0x21b6910, 0xc420c14460, 0xc421f39680, 0x0)
	/Users/aclaus/go/src/github.com/trivago/gollum/core/router.go:63 +0x53c
github.com/trivago/gollum/core.(*SimpleConsumer).directEnqueue(0xc420ba5680, 0xc421f39680)
	/Users/aclaus/go/src/github.com/trivago/gollum/core/simpleconsumer.go:254 +0x1fe
github.com/trivago/gollum/core.(*SimpleConsumer).(github.com/trivago/gollum/core.directEnqueue)-fm(0xc421f39680)
	/Users/aclaus/go/src/github.com/trivago/gollum/core/simpleconsumer.go:92 +0x34
github.com/trivago/gollum/core.(*SimpleConsumer).EnqueueWithMetadata(0xc420ba5680, 0xc422446709, 0x337, 0x6f7, 0xc422420c30)
	/Users/aclaus/go/src/github.com/trivago/gollum/core/simpleconsumer.go:203 +0x82
github.com/trivago/gollum/consumer.(*Kafka).enqueueEvent(0xc420ba5680, 0xc42242c850)
[...]

PR #196, opened 2017-08-03
#195: race condition fix

arnecls:
Some parts of the code neglected the fact that plugins are added during runtime.
The stream- and plugin-registry now use RWMutexes (as they act like caches) for improved read performance.

### DIFF
--- a/core/streamregistry.go
+++ b/core/streamregistry.go
@@ -30,8 +30,8 @@ var (
 type streamRegistry struct {
 	routers     map[MessageStreamID]Router
 	name        map[MessageStreamID]string
-	nameGuard   *sync.Mutex
-	streamGuard *sync.Mutex
+	nameGuard   *sync.RWMutex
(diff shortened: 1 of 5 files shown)


core/streamregistry.go
```diff
@@ -203,28 +191,46 @@ func (registry *streamRegistry) createFallback(streamID MessageStreamID) Router
...
-
-	registry.streamGuard.Lock()
 	registry.routers[streamID] = defaultRouter
-	registry.streamGuard.Unlock()
 
 	CountRouters()
-	if streamID != InvalidStreamID && streamID != WildcardStreamID {
-		CountFallbackRouters()
-	}
+	CountFallbackRouters()
```
msiebeneicher: sure that we want to count "fallback router" incase of the WildcardStreamID?

arnecls: Yes, because you can define a router for wildcard, too.

PR #197, opened 2017-08-03
Make metadata an on-demand field

arnecls:
Profiling showed that adding metadata to a message comes with some overhead.
As it is not needed in many cases, this PR makes it optional.

- Passing metadata to NewMessage will of course use that map directly (unless it is empty)
- GetMetadata creates the map if not existing
- TryGetMetadata returns nil (for read only access)
- FreezeOriginal will take into account that metadata can be nil
- Serialize and Deserialize already supported nil values

### DIFF
--- a/core/message.go
+++ b/core/message.go
@@ -57,9 +57,7 @@ func NewMessage(source MessageSource, data []byte, metadata Metadata, streamID M
 
 	message.data.payload = buffer
 	message.data.streamID = streamID
-	if metadata == nil {
-		message.data.Metadata = make(Metadata)
-	} else {
+	if metadata != nil && len(metadata) > 0 {
 		message.data.Metadata = metadata
 	}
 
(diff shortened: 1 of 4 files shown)


core/message.go
```diff
@@ -194,7 +202,12 @@ func (msg *Message) CloneOriginal() *Message {
 func (msg *Message) FreezeOriginal() {
 	msg.orig.payload = getPayloadCopy(msg.data.payload)
 	msg.orig.streamID = msg.data.streamID
-	msg.orig.Metadata = msg.data.Metadata.Clone()
+
+	if msg.data.Metadata == nil {
```
msiebeneicher: this should be enough:
```go
if msg.data.Metadata != nil {
  msg.orig.Metadata = msg.data.Metadata.Clone()
} 
```

arnecls: well, there is the theoretical case that data.Metadata changes from a map to nil but that should never happen. And the next PR (delayed freeze) will change Metadata behavior anyway.

PR #198, opened 2017-08-03
Move FreezeOriginal to producer side

arnecls:
This PR is part of the performance optimisation round before the release of 0.5.0.

- The streamID has been moved back to core.Message from core.MessageData
- A new field originalStreamID has been added to core.Message
- The originalStreamID is set SetStreamIDAsOriginal is called
- original data in core.Message is now optional
- FreezeOriginal is now only called during core.SimpleProducer.Modulate if the producer contains modulators
- FreezeOriginal can only be called once
- The serialization format has been changed to include all fields
- SerializeOriginal has been removed (because of the previous change)


core/message.go
```diff
@@ -24,16 +24,17 @@ import (
...
 	Metadata Metadata
 }
 
 // Message is a container used for storing the internal state of messages.
 // This struct is passed between consumers and producers.
 type Message struct {
 	data         MessageData
-	orig         MessageData
+	orig         *MessageData
+	streamID     MessageStreamID
```
msiebeneicher: hmm - with three streamIDs we should create an own struct for the stream id handling. But that can also be updated in a later version

core/message.go
```diff
@@ -93,19 +92,37 @@ func (msg *Message) GetPrevStreamID() MessageStreamID {
 
 // GetRouter returns the stream object behind the current StreamID.
 func (msg *Message) GetRouter() Router {
-	return StreamRegistry.GetRouterOrFallback(msg.GetStreamID())
```
msiebeneicher: i would avoid direct access and keep the get method here

core/message.go
```diff
@@ -93,19 +92,37 @@ func (msg *Message) GetPrevStreamID() MessageStreamID {
...
 }
 
 // GetPrevRouter returns the stream object behind the previous StreamID.
 func (msg *Message) GetPrevRouter() Router {
 	return StreamRegistry.GetRouterOrFallback(msg.prevStreamID)
 }
 
+// GetOrigRouter returns the stream object behind the original StreamID.
+func (msg *Message) GetOrigRouter() Router {
+	return StreamRegistry.GetRouterOrFallback(msg.origStreamID)
```
msiebeneicher: what happens if the origStreamID is still `nil`?

core/message.go
```diff
@@ -93,19 +92,37 @@ func (msg *Message) GetPrevStreamID() MessageStreamID {
...
+}
+
 // SetStreamID sets a new stream and stores the current one in the previous
-// stream field.
+// stream field. If the previous stream is invalid and the new stream is not,
+// the new stream will become the original stream.
 func (msg *Message) SetStreamID(streamID MessageStreamID) {
-	msg.prevStreamID = msg.GetStreamID()
-	msg.data.streamID = streamID
+	if streamID != InvalidStreamID && msg.streamID == InvalidStreamID {
```
msiebeneicher: I'm not sure if the following case is not possible: 

filter or formatter set stream id to `InvalidStreamID; router set new stream id to a fallback stream id => in this case original stream id will changed also to the fallback stream id.

core/message.go
```diff
@@ -93,19 +92,37 @@ func (msg *Message) GetPrevStreamID() MessageStreamID {
...
+		msg.origStreamID = streamID
+	}
+	msg.prevStreamID = msg.streamID
+	msg.streamID = streamID
+}
+
+// SetOriginalStreamID acts like SetStreamID but always sets the original stream
+// ID, too. This method should be used before a message is routed for the first
+// time (e.g. in a consumer) or when the original stream should change.
+func (msg *Message) SetOriginalStreamID(streamID MessageStreamID) {
```
msiebeneicher: Function name is wrong - this is a `reset` - I would expect that only the orig stream id will set in a method `SetOriginalStreamID `

core/message.go
```diff
@@ -185,72 +202,92 @@ func (msg *Message) Clone() *Message {
...
 }
 
-func (data MessageData) serialize(prevStreamID MessageStreamID, timestamp time.Time) ([]byte, error) {
+func (data MessageData) serialize(msg *Message) ([]byte, error) {
 	serializable := &SerializedMessage{
-		StreamID:     proto.Uint64(uint64(data.streamID)),
-		PrevStreamID: proto.Uint64(uint64(prevStreamID)),
-		Timestamp:    proto.Int64(timestamp.UnixNano()),
+		StreamID:     proto.Uint64(uint64(msg.streamID)),
+		PrevStreamID: proto.Uint64(uint64(msg.prevStreamID)),
```
msiebeneicher: where is the origStreamID?

core/message.go
```diff
@@ -185,72 +202,92 @@ func (msg *Message) Clone() *Message {
...
 	err := proto.Unmarshal(data, serializable)
 
-	msg := Message{
+	msg := &Message{
 		prevStreamID: MessageStreamID(serializable.GetPrevStreamID()),
 		timestamp:    time.Unix(0, serializable.GetTimestamp()),
 	}
 
-	msg.data.streamID = MessageStreamID(serializable.GetStreamID())
+	msg.streamID = MessageStreamID(serializable.GetStreamID())
```
msiebeneicher: same here: where is the origStreamID?

core/message.go
arnecls: I see no point in doing so. It's not going to be reused and you don't gain any benefit from the separation.

core/message.go
arnecls: it cannot be nil. It can be invalid and in that case a nil router is returned

core/message.go
arnecls: Good point. I'll remove that.

core/message.go
arnecls: SetlStreamIDAsOriginal is propably a better name as it extend SetStreamID by also setting the original

core/message.go
arnecls: Not there as it wasn't there before.
This is a tricky one.
We did not add the original data to serialization when we introduced it as this would have broken deserialization of data written with 0.4.x. In that case we would get data without original data but act as if there was. False assumption = wrong behavior = broken pipelines.

Now looking at that decision in retrostpective I think we should change this.
The pipelines will change drastically anyway, so you should see that you drain any serialized data before switching to 0.5.0 (this should be added to the release notes).
But this should be a PR on its own.

core/message.go
msiebeneicher: no technical benefits - only code quality and -rate increasing (decrease struct properties! ;P

PR #199, opened 2017-08-04
Performance/give consumers option to set meta data

msiebeneicher:
* made meta data setting optional in consumer.file
* made meta data setting optional in consumer.console
* made meta data setting optional in consumer.kafka

### DIFF
--- a/consumer/console.go
+++ b/consumer/console.go
@@ -37,6 +37,8 @@ const (
 //
 // Metadata
 //
+// - NOTE: *The metadata will only set if the parameter `SetMetadata` is active.*
+//
 // - pipe: Name of the pipe the message was received on (set)
 //
 // Parameters
@@ -53,6 +55,11 @@ const (
 // pipe is closed, i.e. when EOF is detected.
 // By default this paramater is set to "true".
 //
+// - SetMetadata: If this value is set to "true" the `Metadata` will set to each message
+// during the creation.
+// By default this parameter is set to "false".
+// * NOTE: This setting needs additional performance.
+//
 // Examples
 //
 // This config reads data from stdin e.g. when starting gollum via unix pipe.
@@ -66,6 +73,7 @@ type Console struct {
 	autoExit            bool   `config:"ExitOnEOF" default:"true"`
 	pipeName            string `config:"Pipe" default:"stdin"`
 	pipePerm            uint32 `config:"Permissions" default:"0644"`
+	hasToSetMetadata    bool   `config:"SetMetadata" default:"false"`
 	pipe                *os.File
 }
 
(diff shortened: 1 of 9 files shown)


consumer/console.go
```diff
@@ -53,6 +55,11 @@ const (
 // pipe is closed, i.e. when EOF is detected.
 // By default this paramater is set to "true".
 //
+// - SetMetadata: If this value is set to "true" the `Metadata` will set to each message
```
arnecls: Better: "When this value is set to true, the fields mentioned in the metadata section will be added to each message. Adding metadata will have a performance impact on systems with high throughput."

Same for the other occurrences of this setting.

consumer/console.go
```diff
@@ -53,6 +55,11 @@ const (
...
+// during the creation.
+// By default this parameter is set to "false".
+// * NOTE: This setting needs additional performance.
```
arnecls: Add this to note as a normal sentence. (see above)

consumer/console.go
```diff
@@ -86,10 +94,14 @@ func (cons *Console) Configure(conf core.PluginConfigReader) {
 
 // Enqueue creates a new message
 func (cons *Console) Enqueue(data []byte) {
-	metaData := core.Metadata{}
-	metaData.SetValue("pipe", []byte(cons.pipeName))
+	if cons.hasToSetMetadata {
+		metaData := core.Metadata{}
```
arnecls:
```go
metaData := core.Metadata {
    pipe: []byte(const.PipeName),
}
```

consumer/file.go
```diff
@@ -97,9 +99,14 @@ const (
 //
 // - PollingDelay: This value defines the duration the consumer waits between
 // checking the source file for new content after hitting the end of file (EOF).
-// The value is in milliseconds (ms). NOTE: this settings only takes effect if
-// the consumer is running in `poll` mode!
+// The value is in milliseconds (ms).
 // By default this parameter is set to "100".
+// * NOTE: This settings only takes effect if the consumer is running in `poll` mode!
```
arnecls: Again: add as a normal sentence

consumer/file.go
```diff
@@ -150,13 +158,17 @@ func (cons *File) Configure(conf core.PluginConfigReader) {
 
 // Enqueue creates a new message
 func (cons *File) Enqueue(data []byte) {
-	metaData := core.Metadata{}
+	if cons.hasToSetMetadata {
+		metaData := core.Metadata{}
```
arnecls: again: shorthand notation

consumer/console.go
msiebeneicher: this is not possible because there are no defined properties in Metadata: 
```go
type Metadata map[string][]byte
```

consumer/file.go
msiebeneicher: same here

PR #204, opened 2017-08-09
Add SkipIfEmpty option to all formatters

arnecls:
## What is the purpose of this pull request

This PR adds an option to all formatters allowing to skip empty payload.
Empty in this case means an empty byte array as well as a non existing metadata field
This is a useful switch when dealing with data that only should be formatted if valid, e.g. data from a kafka topic where some messages have a key set and some haven't.

## Config to verify:

```yaml
In:
    Type: consumer.Console
    Streams: data
    Modulators:
        - format.Timestamp:
            SkipIfEmpty: true
        - format.Envelope

Out:
    Type: producer.Console
    Streams: data
```

## Checklist

- [x] unit test provided
- [x] docs updated

### DIFF
--- a/core/metadata.go
+++ b/core/metadata.go
@@ -22,7 +22,17 @@ func (meta Metadata) SetValue(key string, value []byte) {
 	meta[key] = value
 }
 
-// GetValue returns a meta data value by key
+// TrySetValue sets a key value pair only if the key is already existing
+func (meta Metadata) TrySetValue(key string, value []byte) bool {
+	if _, exists := meta[key]; exists {
+		meta[key] = value
+		return true
+	}
+	return false
+}
+
+// GetValue returns a meta data value by key. This function returns a value if
+// key is not set, too. In that case it will return an empty byte array.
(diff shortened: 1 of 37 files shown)


core/metadata.go
```diff
@@ -41,6 +41,12 @@ func (meta Metadata) Delete(key string) {
 	delete(meta, key)
 }
 
+// HasValue returns true if the given key exists
+func (meta Metadata) HasValue(key string) bool {
```
msiebeneicher: this is a "KeyExists" method - "HasValue" associates that the value is checked also.

arnecls: Maybe we should make this "TryGetValue" as you normally directly want to do something once you know a key actually exists. Thoughts?

ppar: FWIW, I'm more bothered by the fact that ` func (meta Metadata) GetValue(key string) []byte` doesn't distinguish between empty and inexistent values. But changing its behavior now might of course break something.

ppar:
Looks fine to me

PR #205, opened 2017-08-10
Implemented message trace feature

msiebeneicher:

This PR implements a trace message feature with can be used by the option `-tm`.
The PR also introduce an internal `_TRACE_` stream. If producer is registered to `_TRACE_` the message traces will send as messages to this stream. Otherwise the message tracer will dump the traces to StdOut.

```json
{
	"Step": "Enqueued by consumer",
	"PluginID": "StdIn",
	"Payload": "foo",
	"OriginalPayload": "",
	"Stream": "InvalidStream",
	"PrevStream": "",
	"OrigStream": "InvalidStream",
	"Metadata": {
		"bar": "foo",
		"foo": "Zm9v",
		"hostname": "trivago-882.local:foo"
	},
	"Source": "StdIn",
	"Timestamp": "2017-08-10T18:10:20.119179005+02:00",
	"FingerprintID": 2925394050
}

{
	"Step": "Routed",
	"PluginID": "DefaultRouter",
	"Payload": "foo",
	"OriginalPayload": "",
	"Stream": "StreamIn",
	"PrevStream": "InvalidStream",
	"OrigStream": "StreamIn",
	"Metadata": {
		"bar": "foo",
		"foo": "Zm9v",
		"hostname": "trivago-882.local:foo"
	},
	"Source": "StdIn",
	"Timestamp": "2017-08-10T18:10:20.119179005+02:00",
	"FingerprintID": 2925394050
}

{
	"Step": "Enqueued by buffered producer",
	"PluginID": "StdOut",
	"Payload": "foo",
	"OriginalPayload": "",
	"Stream": "StreamIn",
	"PrevStream": "InvalidStream",
	"OrigStream": "StreamIn",
	"Metadata": {
		"bar": "foo",
		"foo": "Zm9v",
		"hostname": "trivago-882.local:foo"
	},
	"Source": "StdIn",
	"Timestamp": "2017-08-10T18:10:20.119179005+02:00",
	"FingerprintID": 2925394050
}
```

## Config to verify

[base test config](https://github.com/trivago/gollum/blob/master/testing/configs/test_formatter.conf)

## Checklist

- [x] unit test provided
- [x] docs updated


flags.go
```diff
@@ -38,6 +38,7 @@ var (
 	flagMemProfile     = tflag.String("pm", "profilemem", "", "Write heap profile results to a given file.")
 	flagProfile        = tflag.Switch("ps", "profilespeed", "Write msg/sec measurements to log.")
 	flagTrace          = tflag.String("tr", "trace", "", "Write trace results to a given file.")
+	flagTraceMessage   = tflag.Switch("tm", "trace-message", "Write message trace results _TRACE_ stream.") //todo: finalise
```
arnecls: change "tr/trace" to "pt/profiletrace" and use "t/trace" for the new flag

core/messagesource.go
```diff
@@ -22,6 +22,9 @@ type MessageSource interface {
 
 	// IsBlocked returns true if the source cannot produce messages
 	IsBlocked() bool
+
+	// GetID returns the pluginID of the message source
+	GetID() string
```
arnecls: As GetID is shared between plugin and message source we should make this an additional interface

makefile
```diff
@@ -142,8 +142,15 @@ coverprofile:
 	@$(BUILD_ENV) go test $(BUILD_FLAGS) -covermode=count -coverprofile=format.cov ./format
 	@$(BUILD_ENV) go test $(BUILD_FLAGS) -covermode=count -coverprofile=filter.cov ./filter
 	@$(BUILD_ENV) go test $(BUILD_FLAGS) -covermode=count -coverprofile=router.cov ./router
+
+	@echo "INFO: start generating profile.cov"
 	@echo "mode: count" > profile.cov
-	@cat ./*.cov | grep -v "mode: " >> profile.cov
+	@cat core.cov | grep -v "mode: " >> profile.cov
```
arnecls: Use find or ls to create an array and loop over that to create the profile.cov.

core/streamregistry.go
```diff
@@ -73,6 +73,9 @@ func (registry streamRegistry) GetStreamName(streamID MessageStreamID) string {
 	case WildcardStreamID:
 		return WildcardStream
 
+	case InvalidStreamID:
```
arnecls: TraceInternalStreamID is missing here

core/messagetracer.go
```diff
@@ -0,0 +1,180 @@
...
+	dump.Timestamp = msg.timestamp
+
+	dump.FingerprintID = mt.createFingerPrintID(&dump)
+
+	return dump
+}
+
+func (mt *messageTracer) createFingerPrintID(dump *messageDump) uint32 {
+	hash := fnv.New32a()
+	hash.Write([]byte(fmt.Sprintf("%s/%s", dump.Source, dump.Timestamp)))
```
arnecls: you can append multiple values to a hash before sume.
Sprintf adds additional overhead which can be avoided.

core/messagesource.go
msiebeneicher: how discussed: because of the reason that formatters are plugins and have no id's we let this change that it is for now.

core/streamregistry.go
msiebeneicher: updated

PR #208, opened 2017-08-15
Case sensitive config (again)

arnecls:

We saw a problem with the conversion from mixed case keys to lower case keys when parsing the configs. In a nutshell we cannot really distinguish between keys that can be case sensitive and those that are not (nested modulator arrays for example).
As of that and because of the fact that unknown config keys will stop gollum from starting we decided to make the config case sensitive again. In addition to that we will now display a suggestion for unknown keys, so that broken configs can be fixed faster.

## Checklist

- [x] docs updated


core/pluginconfig.go
```diff
@@ -87,7 +87,27 @@ func (conf *PluginConfig) registerKey(key string) string {
 	}
 
 	conf.validKeys[path] = true
-	return lowerCaseKey
+	return key
+}
+
+func (conf PluginConfig) suggestKey(searchKey string) string {
+	closestDist := len(searchKey)
+	bestMatch := "something else"
```
msiebeneicher: maybe better to use "no suggestion" or an empty string as default?

PR #211, opened 2017-09-08
Bugfix/fix testconfig exit code

msiebeneicher:

- fixed exit code for successfully config tests
- fixed doc example

## Steps to verify

```sh
gollum -tc config/socket_receive.conf && echo $?
```

## Checklist

- [ ] docs updated

### DIFF
--- a/main.go
+++ b/main.go
@@ -77,9 +77,26 @@ func mainWithExitCode() int {
 	logrus.Debug("GOLLUM STARTING")
 	defer logrus.Debug("GOLLUM STOPPED")
 
-	config := readConfig()
+	configFile := *flagConfigFile
+	if *flagTestConfigFile != "" {
+		configFile = *flagTestConfigFile
+	}
+
+	config := readConfig(configFile)
 	if config == nil {
 		return tos.ExitError // ### exit, config failed to parse ###
+	}
+
+	if *flagTestConfigFile != "" {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -77,11 +77,28 @@ func mainWithExitCode() int {
 	logrus.Debug("GOLLUM STARTING")
 	defer logrus.Debug("GOLLUM STOPPED")
 
-	config := readConfig()
```
arnecls: Why move all this testing and flag check code back to the main function?
The main function should not check for anything but boolean flags. Everything else should go into sub functions.

main.go
```diff
@@ -179,17 +196,6 @@ func readConfig() *core.Config {
...
 
-	if testAndExit {
-		coordinator := NewCoordinator()
-		if err := coordinator.Configure(config); err != nil {
-			logrus.WithError(err).Error("Configure pass failed.")
-		} else {
-			fmt.Println("Config OK.")
-		}
-		coordinator.Shutdown()
-		return nil
```
arnecls: Why not simply exit here instead of returning and all that refactoring?

msiebeneicher: Because an exit in a readConfig() method just feels wrong. 
This PR is more a clean up instead of of refactoring in my point of view.

arnecls:
I adjusted the config file retrieval part a bit.
Would be fine for me with that change.

PR #220, opened 2018-02-03
Changed Docker Images

Fleshgrinder:
* Moved the Dockerfiles to a dedicated directory and renamed them to use a `.docker` extensions, this ensures proper syntax highlighting in editors.
* Added a Dockerfile to build a production-ready image:
  * The base image comes along with some useful scripts.
  * Support for Docker health checks.
  * Support for environment variable replacement upon start-up.
  * Support for required environment variables.
  * Defanged operating system.

It seems as if you are using Docker Hub to build the images, those builds obviously would need to be adjusted to reflect the path changes and to build the new Debian based image as well.

### DIFF
--- a/docker/debian.docker
+++ b/docker/debian.docker
@@ -1,5 +1,5 @@
 FROM fleshgrinder/debian:stretch-go1-build
-COPY . /usr/local/src/github.com/trivago/gollum
+COPY ../ /usr/local/src/github.com/trivago/gollum
 WORKDIR /usr/local/src/github.com/trivago/gollum
 RUN make -sj$(nproc) current
 LABEL com.trivago.image="gollum"


arnecls:
One critique point: Why base everything on custom images?
- If you want scripts, you can easily add them in later step
- Custom images are less trusted
- What about maintenance?

I would propose basing everything on the official debian slim images, everything else LGTM.

docker/debian.docker
```diff
@@ -0,0 +1,25 @@
...
+LABEL com.trivago.image="gollum"
+
+# ------------------------------------------------------------------------------
+
+FROM fleshgrinder/debian:stretch
+ENV HEALTHCHECK_PATH=/_ALL_
+ENV HEALTHCHECK_PORT=8080
+ENV LOG_LEVEL=1
+
+COPY --from=0 /usr/local/src/github.com/trivago/gollum/gollum /usr/bin/gollum
```
arnecls: why do a dual-stage build here but not on alpine?

Fleshgrinder:
Doing it like this would allow us to use the images out of the box in production (and consequently everyone else too). Right now we are building these images on our own, this seems like a waste since everyone else could benefit from this too.

docker/debian.docker
Fleshgrinder: There is nothing in the Alpine image that would benefit from a multistage build. You are not compiling Gollum there on your own. I guess you build Gollum on the host to build that image, right? Probably a very bad idea because I know that your system is BSD based (Mac) and not necessarily compatible with Linux. However, I guess the cross-compile functionality of Go makes this less problematic.

arnecls: Well, you cannot cross-compile anymore if you use cgo (which we by default don't).
There are also some dynamic linker problems between linux distributions that are fixed for now but might resurface in the future.
So it does make sense to build on the distribution you want to run on.
The local-build-and-copy mechanism is from a time where we had not yet encountered the problem so it makes sense to fix it. We just did not, yet.

arnecls:
What exactly prevents you from running this container in production with the import changed to an official debian base image? I don't see anything special besides `/usr/bin/entrypoint` - but could also be something that I simply don't know.

Fleshgrinder:
The most prominent changes compared to the default image are:

- Dedicated user for the service (not root).
- Defanged (no `setuid` or `setguid` binaries).

The `entrypoint` is mainly there to work around a missing feature in Gollum (and most other software): no support for environment variables.

docker/debian.docker
```diff
@@ -0,0 +1,25 @@
...
+RUN mkdir -p /var/spool/gollum && chown -R srv /var/spool/gollum
+
+CMD /usr/bin/entrypoint gollum -c /usr/local/etc/gollum/config.yml -hc ${HEALTHCHECK_PORT} -ll ${LOG_LEVEL}
+USER srv
```
Fleshgrinder: Here is the switch from `root` to a dedicated service user (`srv`).

docker/debian.docker
```diff
@@ -0,0 +1,25 @@
+FROM fleshgrinder/debian:stretch-go1-build
+COPY . /usr/local/src/github.com/trivago/gollum
+WORKDIR /usr/local/src/github.com/trivago/gollum
+RUN make -sj$(nproc) current
+LABEL com.trivago.image="gollum"
+
+# ------------------------------------------------------------------------------
+
+FROM fleshgrinder/debian:stretch
```
Fleshgrinder: This image is already defanged and comes along with a Docker compatible health check script.

arnecls:
After some offline discussion:
1. The base images should be official trivago/* images
2. The changes introduced by these images have to be documented

Once these requirements are met, this PR will be merged.

Fleshgrinder:
Closing because requirement 1. will probably never be satisfied.

Issue #221, opened 2018-02-07
Feature request: Metrics in Prometheus format


dkleuser:
## Issue description

It would be great if Gollums metrics were also exposed in the format that Prometheus expects.

## Context

This would enable us to monitor Gollum with [Prometheus](https://prometheus.io/).

## Possible Solution

There is an official Golang client library for exposing metrics in the required format: https://github.com/prometheus/client_golang
As it is the default that Prometheus expects, it would be nice if the metrics were available at the endpoint `/prometheus`.

Please also see https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels for the allowed characters for metric labels.

arnecls:
Another solution would be to switch to [go-metrics](https://github.com/rcrowley/go-metrics/blob/master/README.md) and use their prometheus adapter: https://github.com/deathowl/go-metrics-prometheus 

However - we should benchmark go-metrics first as metrics already cause quite an overhead in gollum.
Also we must somewhat stay backwards compatible as there are tools around processing gollum metrics ... and we already somewhat broke this with 0.5

PR #222, opened 2018-02-08
Switch metrics from tgo to go-metrics

arnecls:

Exchange tgo.metrics with go-metrics to improve metrics and allow different metrics provider.
Request as by issue #221

## Checklist

- [ ] unit test provided
- [ ] docs updated


core/metrics.go
```diff
@@ -15,199 +15,147 @@
...
+	metricBuild.Update(int64(build))
+	go metrics.CaptureRuntimeMemStats(MetricsRegistry, time.Second)
 }
 
-// CountRouters increases the stream counter by 1
-func CountRouters() {
-	tgo.Metric.Inc(metricRouters)
+// NewSubRegistry creates a new, prefixed metrics registry that can be used
+// to register custom plugin metrics.
+func NewSubRegistry(prefix string) metrics.Registry {
```
ppar: Could NewSubRegistry() and NewPluginRegistry() be renamed to something like NewMetricsSubRegistry() and NewMetricsPluginRegistry()? Currently when you encounter references to these elsewhere in the code, nothing tells you that they're metrics-specific functions (other than the return value's type, which is invisible) - they look like central parts of Gollum's core instead.

ppar:
Looks OK to me - however, I haven't worked closely with a lot of the code involved here.

core/metrics.go
arnecls: good point

arnecls: adjusted. It's now `NewMetricsRegistry` and `NewMetricsRegistryForPlugin`.

arnecls:
merged in order to move forward with 0.6

PR #223, opened 2018-02-20
Doc fix: delimiter description is rendered incorrectly (n instead of \n)

dylanwh:
The delimiter default is \n, but this renders as "n" in restructured text.
\\n produces an \n in the output


andygrunwald:
Thank you Dylan!

PR #225, opened 2018-03-20
commit offsets to kafka when using consumer groups

crewton:

In order to have consumer groups not start from the beginning if they all crash, and to be able monitor where they are; consumers need to commit their offsets. Technically consumer groups work without them, however it can lead to some unwanted behavior in failure scenarios.

## Config to verify

```yaml
kafkaIn:
  Type: consumer.Kafka
  Streams: logs
  Topic: logs
  ClientId: "gollum log reader"
  GroupId: "gollum"
  Servers:
    - "kafka0:9092"
    - "kafka1:9092"
    - "kafka2:9092"
    - "kafka3:9092"
```

## Checklist

- [ ] unit test provided
- [ ] docs updated

### DIFF
--- a/consumer/kafka.go
+++ b/consumer/kafka.go
@@ -397,8 +397,11 @@ func (cons *Kafka) readFromGroup() {
 
 	for !cons.groupClient.Closed() {
 		select {
-		case event := <-consumer.Messages():
-			cons.enqueueEvent(event)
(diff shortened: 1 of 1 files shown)


arnecls:
Thanks for noticing 👍 
This also affects the 0.5.x branch, so I will merge it to that branch,  too.

PR #232, opened 2018-08-22
Multi file scraping

arnecls:

Extend consumer.File to allow parsing multiple files at the same time by using glob patterns.
This introduces breaking changes.

## Config to verify

```yaml
logFilesIn:
    Type: consumer.File
    Streams: logs
    Files: "/var/log/*.log"

logFilesOut:
    Type: producer.Console
    Streams: logs
```

## Checklist

- [x] integration test provided
- [x] docs updated

### DIFF
--- a/consumer/file.go
+++ b/consumer/file.go
@@ -191,8 +191,8 @@ func (cons *File) newObservedFile(name string, stopIfNotExist bool) *observableF
 	case defaultOffset == fileOffsetEnd:
 		cursor.whence = io.SeekEnd
 
-	default:
 	case defaultOffset == fileOffsetStart:
+	default:
 	}
 
 	logger.Info("Starting file scraper")
--- a/consumer/file.go
+++ b/consumer/file.go
@@ -190,6 +190,9 @@ func (cons *File) newObservedFile(name string, stopIfNotExist bool) *observableF
 
 	case defaultOffset == fileOffsetEnd:
 		cursor.whence = io.SeekEnd
+
+	default:
+	case defaultOffset == fileOffsetStart:
 	}
 
 	logger.Info("Starting file scraper")
--- a/consumer/file.go
+++ b/consumer/file.go
@@ -91,7 +91,7 @@ const (
 // By default this parameter is set to "newest".
 //
 // - PollingDelayMs: This value defines the duration in Milliseconds the consumer
-//  waits between checking the source file for new content after hitting the
+// waits between checking the source file for new content after hitting the
 // end of file (EOF). NOTE: This settings only takes effect if the consumer is
 // running in `poll` mode!
 // By default this parameter is set to "100".
--- a/consumer/file.go
+++ b/consumer/file.go
@@ -226,7 +226,7 @@ func (cons *File) observeFile(name string) {
 func (cons *File) observeFiles() {
 	defer cons.WorkerDone()
 
-	if !strings.Contains(cons.fileName, "*") {
+	if !strings.ContainsAny(cons.fileName, "*?") {
 		cons.observeFile(cons.fileName) // blocking
 		return
 	}
--- a/consumer/file.go
+++ b/consumer/file.go
@@ -41,6 +41,11 @@ const (
 	observeModeWatch = "watch"
 )
 
+const (
+	stopIfNotExist  = true
+	retryIfNotExist = false
+)
+
 // File consumer plugin
 //
 // The File consumer reads messages from a file, looking for a customizable
@@ -61,7 +66,7 @@ const (
 // Parameters
 //
 // - File: This value is a mandatory setting and contains the name of the
-// file to read. This filed supports glob patterns.
+// file to read. This field supports glob patterns.
 // If the file pointed to is a symlink, changes to the symlink will be
 // detected. The file will be watched for changes, so active logfiles can
 // be scraped, too.
@@ -90,7 +95,7 @@ const (
 // is defined and the file exists, the DefaultOffset parameter is ignored.
 // By default this parameter is set to "newest".
 //
-// - PollingDelayMs: This value defines the duration in Milliseconds the consumer
+// - PollingDelayMs: This value defines the duration in milliseconds the consumer
 // waits between checking the source file for new content after hitting the
 // end of file (EOF). NOTE: This settings only takes effect if the consumer is
 // running in `poll` mode!
@@ -99,6 +104,10 @@ const (
 // - RetryDelaySec: This value defines the duration in seconds the consumer waits
 // between retries, e.g. after not being able to open a file.
 // By default this parameter is set to "3".
+//
+// - DirScanIntervalSec: Only applies when using globs. This setting will define the
+// interval in secnds in which the glob will be re-evaluated and new files can be
+// scraped. By default this parameter is set to "10".
 //
 // - SetMetadata: When this value is set to "true", the fields mentioned in the metadata
 // section will be added to each message. Adding metadata will have a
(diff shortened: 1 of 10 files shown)


consumer/file.go
```diff
@@ -140,351 +139,118 @@ func init() {
...
-			if err != nil {
-				cons.Logger.Error("Error reading offset file: ", err)
+	offsetFileName := ""
+	defaultOffset := strings.ToLower(cons.defaultOffset)
+	cursor := fileCursor{whence: io.SeekStart}
+
+	switch {
+	case cons.offsetFilePath != "":
+		offsetFileName = fmt.Sprintf("%s/%s.offset", cons.offsetFilePath, filepath.Base(cons.fileName))
+		if offsetFileData, err := ioutil.ReadFile(offsetFileName); err == nil {
```
SHyx0rmZ: This should be `err != nil`.

consumer/fileobservable.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+	default:
+		fs.log.WithError(err).Error("Failed to read file")
+		fs.handle.Close()
+		fs.handle = nil
+		fs.buffer.Reset(0)
+	}
+}
+
+func (fs *observableFile) observePoll(enqueue func([]byte), done chan struct{}) {
```
SHyx0rmZ: It would be nice to use `done <-chan struct{}` instead, so one can easily see that the channel is only to be received from.

consumer/fileobservable.go
```diff
@@ -0,0 +1,226 @@
...
+				"Scraping": actualFileName,
+			})
+		}
+
+		fs.scrape(actualFileName, enqueue, spin.Reset)
+		spin.Yield()
+	}
+}
+
+func (fs *observableFile) observeFSNotify(enqueue func([]byte), done chan struct{}) {
```
SHyx0rmZ: It would be nice to use `done <-chan struct{}` instead, so one can easily see that the channel is only to be received from.

consumer/file.go
```diff
@@ -140,351 +139,118 @@ func init() {
...
+func (cons *File) observeFile(name string) {
+	defer cons.WorkerDone()
+	file := cons.newObservedFile(name)
+	defer file.close()
 
-	for cons.source.state != fileStateDone {
-		cons.read(buffer, sendFunction, spin.Yield, spin.Reset)
-	}
-}
+	enqueue := cons.SimpleConsumer.Enqueue
```
SHyx0rmZ: I do not see an obvious reason to use `cons.SimpleConsumer.Enqueue` here instead of `cons.Enqueue`, when you are using `const.EnqueueWithMetadata` a few lines below. It would be great if you could decide on one style of access for both.

consumer/file.go
```diff
@@ -140,351 +139,118 @@ func init() {
...
+	// TODO: support manual roll again
+	//cons.SetRollCallback(cons.onRoll)
+	cons.SetStopCallback(func() {
+		close(cons.done)
+	})
 
 	// restore default observer mode for invalid config settings
 	if cons.observeMode != observeModePoll && cons.observeMode != observeModeWatch {
-		cons.Logger.WithField("observeMode", cons.observeMode).Errorf("Unknown observe mode '%s'", cons.observeMode)
+		cons.Logger.Warningf("Unknown observe mode '%s'. Using poll", cons.observeMode)
```
SHyx0rmZ: If it's fine for you I'd like to not start the log message with a capital letter, as is idiomatic for error messages.

SHyx0rmZ: I'd also like to do the same for other error/log messages contained in this pull request.

consumer/file.go
```diff
@@ -140,351 +139,118 @@ func init() {
...
-		source.lastStatCheck = time.Now()
+func (cons *File) observeFiles() {
+	defer cons.WorkerDone()
 
-		if newStatErr == nil && oldStatErr == nil && !os.SameFile(newStat, oldStat) {
-			return true
-		}
+	if !strings.ContainsAny(cons.fileName, "*?") {
+		cons.observeFile(cons.fileName) // blocking
+		return
```
SHyx0rmZ: Maybe this special case could be removed to not have an additional branch to think about in the code?

docs/src/gen/consumer/file.rst
```diff
@@ -42,18 +42,20 @@ Parameters
 **File**
 
   This value is a mandatory setting and contains the name of the
-  file to read. The file will be read from beginning to end and the reader
-  will stay attached until the consumer is stopped, so appends to the
-  file will be recognized automatically.
+  file to read. This filed supports glob patterns.
```
SHyx0rmZ: Typo: `This field supports glob patterns.`

docs/src/gen/consumer/file.rst
```diff
@@ -87,16 +89,24 @@ Parameters
...
   
 
-**PollingDelay**
+**PollingDelayMs** (default: 100, unit: ms)
 
-  This value defines the duration the consumer waits between
-  checking the source file for new content after hitting the end of file (EOF).
-  The value is in milliseconds (ms). NOTE: This settings only takes effect if the consumer
-  is running in `poll` mode!
+  This value defines the duration in Milliseconds the consumer
```
SHyx0rmZ: Typo: `milliseconds`

docs/src/gen/consumer/file.rst
```diff
@@ -87,16 +89,24 @@ Parameters
...
+  running in `poll` mode!
   By default this parameter is set to "100".
   
   
 
+**RetryDelaySec** (default: 3, unit: s)
+
+  This value defines the duration in seconds the consumer waits
+  between retries, e.g. after not being able to open a file.
+  By default this parameter is set to "3".
```
SHyx0rmZ: Is there a reason that values are not written as `code`? It would also be nice, to always make sure that the correct value is written inside the documentation, especially if there's a difference between `"3"` and `3`.

consumer/file.go
```diff
@@ -98,12 +90,16 @@ const (
 // is defined and the file exists, the DefaultOffset parameter is ignored.
 // By default this parameter is set to "newest".
 //
-// - PollingDelay: This value defines the duration the consumer waits between
-// checking the source file for new content after hitting the end of file (EOF).
-// The value is in milliseconds (ms). NOTE: This settings only takes effect if the consumer
-// is running in `poll` mode!
+// - PollingDelayMs: This value defines the duration in Milliseconds the consumer
```
SHyx0rmZ: Typo: `milliseconds`

docs/src/gen/consumer/file.rst
arnecls: that's in line with the rest of the documentation, i.e. a change would be beyond this PR

consumer/file.go
arnecls: Actually not, because single file scraping is a lot simpler (the next commit will make this even clearer).

docs/src/gen/consumer/file.rst
SHyx0rmZ: I'm just wondering if, considering it's already inconsistent in just this single file, we should start making it consistent bit by bit while accepting inconsistency between the consistent files and the inconsistent files in the meantime. If you think that's not the right thing to do I'm also fine with leaving it the way it currently is, I just wanted us to at least consider a possible cleanup.

docs/src/gen/consumer/file.rst
arnecls: fixed (line moved)

docs/src/gen/consumer/file.rst
arnecls: fixed

docs/src/gen/consumer/file.rst
arnecls: If it's inconsistent in more files (I was not aware of that) my previous comment is even more true. We should find a consent here.

consumer/file.go
arnecls: Most log messages in Gollum start with an uppercase letter.
Is this really idiomatic? Gofmt complains on a `.` at the end of a message, but I don't know of any other convention (source?).

SHyx0rmZ: https://github.com/golang/go/wiki/Errors#user-content-errors

arnecls: :+1: but again - as this has not been done throughout the code it should be a cleanup PR by itself (after #231 has been merged please).

PR #239, opened 2018-11-29
Enable plugin version injection in makefile

M3t0r:

Make it easy to add the version of compiled in plugins to the gollum version string.

## Example to verify

```
make build GOLLUM_RELEASE_SUFFIX=pkgVersionSuffix
./gollum -version
```

## Checklist

- [ ] unit test provided
- [x] docs updated

### DIFF
--- a/makefile
+++ b/makefile
@@ -3,7 +3,7 @@
 GOLLUM_TAG := $(shell git describe --always --tags --match "v*" | sed -E 's/^v([0-9\.]+)(-[0-9]+){0,1}((-)g([a-f0-9]+)){0,1}.*/\1\2\4\5/')
 GOLLUM_RELEASE_SUFFIX := # Can be set for custom releases to also include plugin versions
 GOLLUM_DIRTY := $(if $(shell git status --porcelain),-dirty)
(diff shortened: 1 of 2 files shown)


makefile
```diff
@@ -1,8 +1,9 @@
 .DEFAULT_GOAL := help
 
 GOLLUM_TAG := $(shell git describe --always --tags --match "v*" | sed -E 's/^v([0-9\.]+)(-[0-9]+){0,1}((-)g([a-f0-9]+)){0,1}.*/\1\2\4\5/')
+GOLLUM_RELEASE_SUFFIX := # Can be set for custom releases to also include plugin versions
 GOLLUM_DIRTY := $(if $(shell git status --porcelain),-dirty)
-GOLLUM_VERSION := $(join $(GOLLUM_TAG),$(GOLLUM_DIRTY))
+GOLLUM_VERSION := $(GOLLUM_TAG)$(GOLLUM_DIRTY)$(if $(GOLLUM_RELEASE_SUFFIX),-$(GOLLUM_RELEASE_SUFFIX))
```
arnecls: To keep the version string semver compatible, we should treat the suffix as metadata, see https://semver.org/spec/v2.0.0.html#spec-item-10 .
In this case it should be `+$(GOLLUM_RELEASE_SUFFIX)`.

PR #247, opened 2019-04-12
Fix empty metadata after message cloning

msiebeneicher:

- fixed cloning metadata
- fixed unit test for cloning metadata

## Checklist

- [x] unit test provided


arnecls:
Found and fixed an issue after merging.
The thing with CloneOriginal is that we create a copy of the current state first and then either need to clone the current metadata (to avoid changes "leaking" into the copy) or explicitly set it to nil as the original (as defined by the "freeze" operation) does not have metadata.

See commit: https://github.com/trivago/gollum/commit/7c14a24d471b75c5b1642f2a3a961b7028380c3c

PR #251, opened 2019-04-24
Delete failing errors_test.go

M3t0r:

The only test case in that file only tests if the constructor function exists and returns an error with a correctly formatted string. This format changed in [4ce38f3546eb88b7265acb8d02e2e8f37f5fc5eb](https://github.com/trivago/gollum/commit/4ce38f3546eb88b7265acb8d02e2e8f37f5fc5eb#diff-3c5a3f4e4c920721cbc650a3e4d5eec4L36) but wasn't updated in the test. And everything besides the format is also tested by the go compiler at built time. We don't need a special test for that.

## Checklist

- [ ] unit test provided
- [ ] docs updated


arnecls:
in 0.6.0 we still have this test, but I agree that it is somewhat pointless.

M3t0r:
But it [still passes in 0.6](https://travis-ci.org/trivago/gollum/jobs/524638088#L515). Even though it's pointless :-)

arnecls:
because I fixed the error message there :D

PR #254, opened 2019-05-17
Add permission support to syslog sockets

### LINKED ISSUE #253: Syslogd consumer should support setting file permissions
While the socket consumer [supports setting socket permissions](https://github.com/trivago/gollum/blob/4ce38f3546eb88b7265acb8d02e2e8f37f5fc5eb/consumer/socket.go#L50-L52), the syslogd consumer does not.

## Issue description

When starting a syslogd consumer, the socket gets created with the same permissions as the gollum process. Sometimes this can be a problem; for example when another process  wants to write into the socket which runs with insufficient permissions.

## Context

More concretely, we have an NGINX webserver, which spawns workers as `nobody` while gollum runs as a dedicated user. As a result, the workers cannot log into the socket.

## Possible Solution

It would be helpful to be able to set the permissions in the syslogd consumer similar to the socket consumer. We could port over the code almost verbatim.

## Steps to Reproduce (for bugs)

1. Start gollum with a syslogd consumer.
2. Try to log into the newly created socket with an unprivileged user.

### Used versions

* gollum version: 0.5.4
* go version: 1.12
* Operating System and version: macOS 10.14.1

mre:
This fixes #253.
The default permissions are "0770" just like the socket consumer.

## Config to verify

```yaml
SyslogdSocketConsumer:
    Type: consumer.Syslogd
    Streams: syslog
    Address: "unix:///dev/log"
    Format: "RFC3164"
    Permissions: "0540"

"StdOut":
    Type: "producer.Console"
    Streams: syslog
```

## Checklist

- [X] `make test` executed successfully
- [ ] unit test provided
- [X] integration test provided
- [X] docs updated

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,42 +2,55 @@
 
 
 [[projects]]
+  digest = "1:355a000a06dc127c555e408c3778f9968fe4c1680ea6a1c13ba5612274b78551"
   name = "github.com/Shopify/sarama"
   packages = ["."]
+  pruneopts = ""
   revision = "3b1b38866a79f06deddf0487d5c27ba0697ccd65"
   version = "v1.15.0"
 
 [[projects]]
+  digest = "1:61e512f75ec00c5d0e33e1f503fbad293f5850b76a8f62f6035097c8c436315d"
   name = "github.com/abbot/go-http-auth"
   packages = ["."]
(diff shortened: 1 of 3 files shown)


arnecls:
Thank you. I will cherry pick this into master

PR #256, opened 2019-07-18
Respect cgroup limits

arnecls:

When not using the -n flag, gollum sets GOMAXPROCS to the number of CPUs available on the machine. When running in docker containers this can lead to performance problems when cpu-limits are enforced via cgroups.

This PR will set the cgroup applied limit (if set and possible) when -n is not used.

## Checklist


arnecls:
tested working in production

